PR #5, opened 2015-07-21
Fixed Broken Stackoverflow Go Link

callmehiphop:
The gcloud-go link was mislabeled as gcloud-java causing the markdown to render incorrectly as well as direct the user to incorrect content.


stephenplusplus:
Good catch. Thanks!

PR #7, opened 2015-07-22
Make H3s - try webhook again - do not merge

ryanseys:


ryanseys:
Ok cool. I was getting 403s after a little while with the webhook. Setting up a secret using internal tool fixed this. Should be all good to go now. Sorry for the PR spam.

stephenplusplus:
WE DON'T WANT H3S, RYAN.

ryanseys:
Fine then!

PR #13, opened 2015-08-03
docs: added hyphens to shell snippet vars

callmehiphop:
`gcloud-node` injects this file into the dom, but all verbiage after the tag name (e.g. `name` in `<project name>`) gets converted to an HTML attribute and ends up rendering like this `<project name="">`. 

Adding some hyphens to work around that.


stephenplusplus:
Crazy Angular.

PR #16, opened 2015-08-21
Add Scopes for Compute Engine & DNS

stephenplusplus:


authorization/readme.md
```diff
@@ -8,9 +8,14 @@ If you are running this client on Google Compute Engine, we handle authorization
...
   * **BigQuery**
     * `https://www.googleapis.com/auth/bigquery`
+  * **Compute Engine**
+    * `https://www.googleapis.com/auth/compute`
   * **Datastore**
     * `https://www.googleapis.com/auth/datastore`
     * `https://www.googleapis.com/auth/userinfo.email`
+  * **DNS**
+    * `https://www.googleapis.com/auth/ndev.clouddns.readwrite`
+    * `https://www.googleapis.com/auth/cloud-platform`
```
dhermes: Put this somewhere more generic.

Maybe comment on all the services covered by this scope?

stephenplusplus: I'm actually getting this list from the scopes we use in node. All of these are exact copies of the scopes we give; no more or less. I wonder if we don't need `cloud-platform` for DNS, and only the first one. And maybe we never need `cloud-platform`? I don't remember where I got that from in the first place :confused:

dhermes: It's listed in the discovery document for the service probably, but it's a very broad scope (like all of the cloud ones).

stephenplusplus: Gotcha, thanks for pointing that out. Lmkwyt of the updates I just pushed.

dhermes:
LGTM

stephenplusplus:
Thanks for the review!

PR #23, opened 2015-08-24
Add code of conduct

blowmage:
This file should be added to the root directory of gcloud projects.
Follow AngularJS's lead by adapting the Contributor Covenant.

### DIFF
--- a/conduct/CODE_OF_CONDUCT.md
+++ b/conduct/CODE_OF_CONDUCT.md
@@ -1,0 +1,22 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free experience for everyone, regardless of level of experience, gender, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, or nationality.
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery
+* Personal attacks
+* Trolling or insulting/derogatory comments
+* Public or private harassment
+* Publishing other's private information, such as physical or electronic addresses, without explicit permission
+* Other unethical or unprofessional conduct.
+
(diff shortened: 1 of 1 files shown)


quartzmo:
Other documents follow a convention of `<root_dir>/<topic_name>/readme.md`. For example: `troubleshooting/readme.md`. I think this may have to do with the Angular application used to dynamically include content, although I'm not absolutely certain of the origin of this convention.

stephenplusplus:
I only did that to group each logically, thinking they might each have their own images or related documents.

blowmage:
The convention I've seen on other projects is that the file is in the root of the repo, similar to README. This is where I believe folks will look for and expect to see it.

stephenplusplus:
Makes sense to me. Do we need both formats?

blowmage:
I added both formats here for completeness, but each gcloud project should pick the format that makes the most sense. If you aren't already using markdown then you can choose the plain text format.

stephenplusplus:
I'd rather just have the markdown file, since the rest of our repo is just markdown. Each gcloud project is GH hosted (I think?), so markdown should be ok.

blowmage:
Removed plain text version.

stephenplusplus:
Cool, thanks. I suppose leave it up to @jgeewax to merge?

blowmage:
Once this PR is accepted I'll send PRs to all the gcloud projects to add the file.

conduct/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,22 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities.
```
jgeewax: Any chance we can follow the semantic line break style? http://rhodesmill.org/brandon/2012/one-sentence-per-line/

jgeewax:
Content :+1:

blowmage:
How about wrapping at 80 characters?

jgeewax:
80-characters kind of misses the point of the semantic breaks... The goal is to have a break be "at the end of a clause" (code already has this -- for the most part).

blowmage:
Can someone take a stab at adding semantic breaks to this document. My attempts so far have been uneven and I feel have hurt the readability. /cc @quartzmo

PR #26, opened 2015-08-27
Adding some optional / reduced scopes to auth doc.

dhermes:
Also remove the email scope from datastore. It's in the discovery doc but doesn't actually work.

@pcostell Can you confirm that `email` scope doesn't work? I noticed it's not even included in https://www.googleapis.com/discovery/v1/apis/datastore/v1beta3/rest

----

Sources for updating scope lists:

https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest
https://www.googleapis.com/discovery/v1/apis/datastore/v1beta2/rest
https://www.googleapis.com/discovery/v1/apis/datastore/v1beta3/rest
https://www.googleapis.com/discovery/v1/apis/pubsub/v1/rest
https://www.googleapis.com/discovery/v1/apis/storage/v1/rest

### DIFF
--- a/authorization/readme.md
+++ b/authorization/readme.md
@@ -8,9 +8,11 @@ If you are running this client on Google Compute Engine, we handle authorization
 
   * **All APIs**
     * `https://www.googleapis.com/auth/cloud-platform`
+    * `https://www.googleapis.com/auth/cloud-platform.read-only`
 
   * **BigQuery**
     * `https://www.googleapis.com/auth/bigquery`
+    * `https://www.googleapis.com/auth/bigquery.insertdata`
   * **Compute Engine**
     * `https://www.googleapis.com/auth/compute`
   * **Datastore**
@@ -25,6 +27,8 @@ If you are running this client on Google Compute Engine, we handle authorization
     * `https://www.googleapis.com/auth/userinfo.email`
   * **Storage**
     * `https://www.googleapis.com/auth/devstorage.full_control`
+    * `https://www.googleapis.com/auth/devstorage.read_only`
+    * `https://www.googleapis.com/auth/devstorage.read_write`
 
 ## On Your Own Server
 


stephenplusplus:
Are the addictions less privileged scopes? If a user lists all of them when creating a VM, I wonder what one would win; most privileged or last one selected?

Ignore me if this doesn't make sense.

dhermes:
Addictions! Funny typo.

I don't know the answer but I'd guess it's most privileged. We could always just check. Maybe @jgeewax knows off the top of his head?

blowmage:
Can we trust this info on Datastore v1beta3? Is there a target date for when that will be available?

I don't see any mention of defining multiple Datasets associated to a Project (ala BigQuery) in the [v1beta3 discovery doc](https://www.googleapis.com/discovery/v1/apis/datastore/v1beta3/rest). Is that coming in v1beta3?

pcostell:
Yes, v1beta3 will require either datastore or cloud-platform. However I believe v1beta2 still requires email (in addition to datastore). I was under the impression that there weren't really more vs. less privileged scopes. i.e. for Datastore, cloud-platform "covers" datastore only because we've explicitly allowed either.

dhermes:
@pcostell So you're saying `https://www.googleapis.com/auth/datastore` on its own is insufficient?

---

I can (sort-of) confirm, I just commented out our [scopes][1] (which had both) and just used the datastore scope without email. The result was a `Unauthorized: 401 Invalid Credentials` exception.

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/blob/2c5df509433e76413f926755ac535642ed6467ea/gcloud/datastore/connection.py#L25-L26

pcostell:
For v1beta2, I believe just datastore in insufficient. For v1beta3, just
datastore OR just cloud-platform is sufficient (and it should work if you
have both).

stephenplusplus:
LGTM.

PR #29, opened 2015-09-16
Changing uses of authorization to authentication.

dhermes:


dhermes:
@stephenplusplus PTAL

stephenplusplus:
Thanks for doing this, I feel much better having this in your hands!

dhermes:
:+1:

PR #34, opened 2015-11-19
Shared docs

### LINKED ISSUE #27: Make it possible to get back to the landing page?
If I get to the inner page, getting back to the landing page means I have to delete some text in the URL... Kind of annoying.

Can we maybe make the GCP logo (top top left) take us to the landing page, and the "gcloud" text take us to the top level of the inner docs?

topleft

Who | for language | Vote
---|---|---
@okdave|Go|n/a (no landing page)
@aozarov|Java|n/a (no inner page)
@mziccard|Java|n/a (no inner page)
@stephenplusplus|Node|(Logo to homepage, text to doc landing page)
@callmehiphop|Node| Logo to homepage, text to doc landing page
@dhermes|Python|
@tseaver|Python|
@blowmage|Ruby| Both link to homepage
@quartzmo|Ruby| Logo to homepage, text to doc landing page

### LINKED ISSUE #33: Share HTML/CSS across docsets
Continue discussing in the link below... I thought we just needed a link/reminder here, since it just took me several minutes to dig it up :)

https://github.com/GoogleCloudPlatform/gcloud-python/issues/1094

### LINKED ISSUE #42: docs: where will each gcloud library put their docs JSON?
Carrying on from https://github.com/GoogleCloudPlatform/gcloud-common/pull/34#issuecomment-162110726

The docs being pushed to each repo will happen automatically, so the task of building the docs that match your release pace will happen in your repo.

I think our Angular app will need to know where to look for a JSON file that serves as a "table of contents". That file would just define the versions available for your project, and possibly the name of the latest version:

```json
{
  "versions": [
    "v0.1.0",
    "v0.1.1"
  ],
  "latest": "v0.1.1"
}
```

Then if we place our JSON files inside of gh-pages/versions/__versioname__.json, the app would know where to load from.

Does this work?

### LINKED ISSUE #43: docs: handle custom docs
Our docs template can pull in our gcloud-common markdown files (the ones that live in this repo, such as Authentication, Troubleshooting, etc). Will we be able to support loading custom documents, and if so, how will our libraries be able to do that?

### LINKED ISSUE #44: docs: landing page
Each of our landing pages are distinct, including different descriptions, examples, and getting started instructions. In our generic docs template, can we support all of these being customized without duplicating templates/styles, etc?

### LINKED ISSUE #46: docs: linking to custom types
I think one of the trickier things we still need to solve is linking to custom types. As it is today we denote custom types with special syntax (e.g. `{module:bigquery#query}`) in [our JSDoc](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/bigquery/index.js#L116). This has worked well for us, but I don't think it's something that's going to transition to other languages very well, so I believe we need to come up with a new way of accomplishing this.

My initial thought is to just modify the `returns` array to look something like this (although I have no idea what the effort would be to implement this in python/ruby/php/etc.)

```javascript
{
  "returns": [{
    "types": [
      // use a string for plain built in types..
      "string",
      // and object notation for custom types
      {
        "service": "bigquery",
        "page": "table" // page would be optional?
      }
    ]
  }]
}
```

Thoughts?

callmehiphop:
[Demo](http://callmehiphop.github.io/gcloud-node/#/)
[Code](https://github.com/callmehiphop/gcloud-node/tree/gh-pages)

Closes #27 
Closes #33
Closes #42
Closes #43 
Closes #44
Closes #46 

#### Todos

- [x] Finish JSON schema - #33 
- [x] Transform gcloud-node site into templated version - https://github.com/GoogleCloudPlatform/gcloud-node/pull/1098
- [x] Generate all JSON for gcloud-node - https://github.com/GoogleCloudPlatform/gcloud-node/pull/1098
- [x] Mobile version switcher
- [x] Mobile menu containing sitemap

---

#### Installing

I've pushed a (mostly) finished version of the docs site up. To run locally requires a few steps.

1 - Run the following commands (or skip them)

```
$ git@github.com:GoogleCloudPlatform/gcloud-common.git
$ cd gcloud-common/site
$ npm install & bower install
```

2 -  Create a [`manifest.json` file](https://gist.github.com/callmehiphop/3271d86686924c62b25f) - click for an example manifest with comments. This should end up in the `src` folder
3 - Generate content that maps to the URLs within your manifest file
4 - Run the following

```
$ gulp serve
```

5  - Yell at me if that didn't work :cry:

#### Deploy Strategy
I believe deploys will work like something like this..

1 - gcloud-common is cloned and built
2 - the docs site is then copied to the gcloud-* gh-pages branch
3 - the manifest.json from the gcloud-* repo is put in the gh-pages root directory
4 - a folder containing all of the content (json/markdown/html/etc.) will also be put in the root directory

#### Content Hierarchy

```
json
├─── home.html // landing page not version dependent
└─── v0.27.0 // folder per release
      ├─── contributing.md // versioned markdown/html files
      └──── bigquery // service
             ├─── index.json // index.json in this case would be the bigquery overview
             └──── table.json // table.json would be bigquery -> table overview

manifest.json
```

So rather than a file for each version, there's a folder. My thoughts are this - guides might change as versions change and it could be pretty bad to provide the user with new authentication guides for older versions of the library since it could have changed.

#### Notable changes from previous gcloud-node site

* Using `gulp` for a better workflow
* Migrated from standard angular router to angular-ui-router
* Moved client side dependencies to bower components
* Refactored to remove dead code and to follow [community styleguide](https://github.com/johnpapa/angular-styleguide)

#### Site Hierarchy

```
Landing Page
└─── Docs (gcloud default page)
      ├──── Guides (Getting Started section) 
      └──── Services (API section)
```

#### Language Hooks

There are a couple of places where I decided to create a hook of sorts to allow for custom markup per language.

* [Custom Type](https://github.com/callmehiphop/gcloud-common/blob/shared-docs/site/src/app/components/custom-type/custom-type.directive.js) - In order to identify and link to custom types within return and param blocks, I had my generated JSON spit out anchor tags with an attribute called `custom-type`. Then in the app there is a directive that will parse that attribute and allow one to build a url based on the attribute value.
* [API Page Headers](https://github.com/callmehiphop/gcloud-common/blob/shared-docs/site/src/app/components/page-header/page-header.directive.js) - It looked like each portal used a different style for an API overview page - but as far as I could tell they are just the friendly names of the API Overviews. So for this one an array containing the friendly names is provided allowing them to prefixed/joined/etc. in whatever way is fit.


stephenplusplus:
Can this work to replace the deploy script?

```
gcloud-node/
  gh-pages/
    versions/
      0.10.0.json
      ...
      0.25.0.json
    versions.txt
    index.html
```

`index.html`
```html
<script src="https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-common/master/src/app.js"></script>
```

The idea being, `index.html` only acts as a bootstrapper that pulls from the latest gcloud-common docs site app.js. That app.js would be responsible for loading views, reading from versions.txt and the matching JSON files from a path relative to the library that includes it.

callmehiphop:
Can you confirm that works? I've read about people running into issues trying to use raw.githubusercontent.com as a cdn due to incorrect mime types.

stephenplusplus:
Good call. There's rawgit, though it has a catch: https://rawgit.com/faq#no-uptime-guarantee

> **I need guaranteed 100% uptime. Should I use cdn.rawgit.com?**

> Probably not.

> RawGit is a free service and cannot provide any uptime or support guarantees, even for cdn.rawgit.com. While I do my best to keep things running, things sometimes go wrong. Sometimes there are network or provider issues outside my control, sometimes abusive traffic temporarily affects response times, and sometimes I break things (although I try really hard not to).

> Since I run RawGit in my spare time, with my own money and with CDN hosting generously donated by MaxCDN, it has a budget that's probably less than you pay for coffee in a given month. My goal is to help other open source developers get their projects up and running, but if you need to serve files that are crucial to your business, you should probably pay for a host with well-funded infrastructure and uptime guarantees.

@jgeewax should we avoid the hook/push process and use rawgit, or is it too risky?

dwsupplee:
What about just letting gcloud-* host the full landing page template?

You should be able to link out to raw.githubusercontent.com as long as we enforce the correct type:

```
<script type="text/javascript" src="https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-common/master/src/app.js"></script>
```

stephenplusplus:
> You should be able to link out to raw.githubusercontent.com as long as we enforce the correct type:

We can't control it, they return an HTTP header which restricts the file from being executable. But this is also harder, because any files that are loaded after that file (directives, filters, controllers, etc) will face the same restrictions.

dwsupplee:
Bah, yep. Outside of hosting our own proxy something like rawgit might be the only option here then.

callmehiphop:
IMO if there's a risk and we can easily circumvent it, which in this case I believe we can, then we should.

dwsupplee:
Yeah, I would agree. My vote goes for not trying to hotlink out to the raw file.

jgeewax:
> @jgeewax should we avoid the hook/push process and use rawgit, or is it too risky?

I just took a look at the deploy script -- it doesn't seem that complicated. I think it might be better to just push copies of code out to each repo (ie, from gcloud-common -> gcloud-node) and continuing as usual. Cool?

stephenplusplus:
:+1:

quartzmo:
@callmehiphop Thanks for the link in gcloud-ruby 454. Is the plan still (per @jgeewax's [comment above](https://github.com/GoogleCloudPlatform/gcloud-common/pull/34#issuecomment-158391433)) to distribute the Angular app and other assets for reuse in each gcloud project? Any clue how?

stephenplusplus:
@quartzmo https://github.com/GoogleCloudPlatform/gcloud-common/pull/35 will do the deploying to our gcloud-* gh-pages branches on every merge here.

quartzmo:
@stephenplusplus So just to clarify: For gcloud-ruby to publish new docs for a release, its build will push updated JSON-format docs to `src` in gcloud-common? Do I have this right?

stephenplusplus:
Oh sorry, there are so many angles to this :dizzy_face:. I think we still need to figure that out. Opened a new issue: https://github.com/GoogleCloudPlatform/gcloud-common/issues/42

callmehiphop:
Pushed a large update, I need to push an example of this working as well as the JSON specific materials to gcloud-node. I plan on doing this shortly.

stephenplusplus:
Looks amazing. :star: :rainbow: :sun_with_face: 

Can we switch the `custom-type` attr to `data-custom-type` just for maximum spec compliancy?Also, React [might not like](https://facebook.github.io/react/docs/jsx-gotchas.html#custom-html-attributes) the shorthand, should there be a day we pull our docs away from Angular (or make this whole docs system available for public use).

stephenplusplus:
> Edit on Github functionality - this was more intentional since the getting started sections now contain markdown from multiple sources, I found that button to be misleading. Feedback around this would appreciated.

I think we need it in some form. On a page that combines multiple sources, maybe just link to the most custom page? Or maybe we can re-think how we present the "Edit" functionality, and have a button go above/beside/atop each section so the link goes to the right place always?

callmehiphop:
> Or maybe we can re-think how we present the "Edit" functionality, and have a button go above/beside/atop each section so the link goes to the right place always?

This was my initial thought, but I didn't really have any great ideas.

stephenplusplus:
Maybe just a little on-hover plain pencil icon that appears in the upper right corner of the text container?

callmehiphop:
> Maybe just a little on-hover plain pencil icon that appears in the upper right corner of the text container?

That could work, I'll play around with that.

stephenplusplus:
@callmehiphop please ping when there are changes we can look at.

Also, is there a way to set up a runnable demo that we can see, maybe using some sample JSON from node?

callmehiphop:
@stephenplusplus sorry about that, I pushed a [demo](http://callmehiphop.github.io/gcloud-node/#/) and added links in the original post.

stephenplusplus:
For this todo:

> Create script to deploy site to gh-pages - PR sent #35 (needs revising)

Can you drop a quick note in #35 about what I can change?

stephenplusplus:
I noticed the nav is missing from the hamburger menu, which now only flips between languages. Was that intentional? Having the only way to move through the site at the absolute bottom of the page isn't ideal. (https://github.com/GoogleCloudPlatform/gcloud-node/pull/678, https://github.com/GoogleCloudPlatform/gcloud-node/pull/688)

stephenplusplus:
More generally, for all of the missing features you listed:

> - Mobile version switcher
> - Mobile menu containing sitemap
> - Edit on Github functionality - this was more intentional since the getting started sections now contain markdown from multiple sources, I found that button to be misleading. Feedback around this would appreciated.

Should we add those to the To Dos, so that we don't merge the PR without them?

stephenplusplus:
> Pictured: Dave from three months ago with Dave from right now.

:clap: Nice work! :tada:

callmehiphop:
> Should we add those to the To Dos, so that we don't merge the PR without them?

Good idea, I'll get on it!

callmehiphop:
> Pictured: Dave from three months ago with Dave from right now.

I'm glad that I'm still stylin' after all this time!

callmehiphop:
**Update:** Latest demo link has most recent (and complete) doc site and JSON schema as discussed per https://github.com/GoogleCloudPlatform/gcloud-common/issues/42 and https://github.com/GoogleCloudPlatform/gcloud-node/issues/1084

site/src/app/service/service.controller.js
```diff
@@ -0,0 +1,47 @@
...
+      }
+
+      if (b.metadata.constructor) {
+        return 1;
+      }
+
+      var aName = a.metadata.name;
+      var bName = b.metadata.name;
+
+      return +(aName > bName) || +(aName === bName) - 1;
```
stephenplusplus: What's this fanciness do?

site/src/app/service/service.html
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+    <section ng-if="method.params.length">
+      <h4>Parameters</h4>
+      <table class="table">
+        <tbody>
+          <tr ng-repeat="param in method.params" ng-class="{ 'param-optional': param.optional, 'param-nullable': param.nullable }">
+            <th scope="row" class="param">
+              <span ng-if="param.parent" class="param-parent">
+                <div>{{::param.parent}}</div>
+                &#8627;
```
stephenplusplus: Should be able to just use ↳

site/src/index.html
```diff
@@ -0,0 +1,66 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>gcloud</title>
+    <meta name="description" content="">
```
stephenplusplus: Should we put something basic here: "Documentation for Google Cloud Platform's client library."?

site/src/app/service/service.controller.js
callmehiphop: Sorts all service methods so that if a constructor is present, it is the first to be displayed and then all other methods are alphabetical.

stephenplusplus: Sorry for vagueness, but I meant the `+() || +() -1`

site/src/app/service/service.html
callmehiphop: Sure, I think I just copied this from gcloud-node.

site/src/index.html
callmehiphop: Ah, yes, I probably skipped this because I wasn't sure what to go with..

site/src/app/service/service.controller.js
callmehiphop: Ahh, ok. I'm typecasting booleans to numbers (shorthand) to achieve the following

```javascript
if (aName > bName) return 1;
if (aName === bName) return 0;
return -1;
```

If you think it's too confusing, I don't mind ditching it

stephenplusplus: It's cool, you maniac.

callmehiphop: :goberserk:

PR #35, opened 2015-11-19
Add a build script to deploy docs to each client library

stephenplusplus:
This script should deploy the src/ directory to each repo after every merge to master.

I'm not a good basher, so maybe there's a way to only do this when a push to master actually involved a change to the src/ directory?

### DIFF
--- a/deploy-docs.sh
+++ b/deploy-docs.sh
@@ -1,0 +1,51 @@
+#/bin/bash
+
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ev
+
+# This script takes the doc contents that exist under the `src` directory and
+# pushes them to a the `src` directory in the gh-pages branch of the target
+# repos (gcloud-node, gcloud-php, etc...)
+
+# Only execute this script if this is a push to master.
+if [ "${TRAVIS_BRANCH}" != "master" ] || [ "${TRAVIS_PULL_REQUEST}" != "false" ]
+  then exit 0
+fi
+
+function deploy_docs {
+  git submodule add -f -b gh-pages https://${GH_OAUTH_TOKEN}@github.com/$1 gh-pages
+
+  # If there is an existing `src` directory in the destination's gh-pages
+  # branch, remove it (we're going to replace it).
+  if [ -d gh-pages/src ]
+    then git rm -rf gh-pages/src
+  fi
+
+  git rm index.html
+
+  gulp build
+  cp -R dist/* gh-pages
+  cd gh-pages
+  git add src
+  git add index.html
+  git config user.name "selfiebot"
+  git commit -m "Pushing a doc site update
+
+  https://github.com/GoogleCloudPlatform/gcloud-common/commit/${TRAVIS_COMMIT}"
+  git push https://${GH_OAUTH_TOKEN}@github.com/$1 HEAD:gh-pages
+}
+
+deploy_docs "googlecloudplatform/gcloud-node"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+script: ./deploy-docs.sh


deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+  fi
+
+  cp -R src gh-pages
+  cd gh-pages
+  git add src
+  git config user.name "selfiebot"
+  git commit -m "Update docs for ${TRAVIS_TAG}"
+  git push https://${GH_OAUTH_TOKEN}@github.com/$1 HEAD:gh-pages
+
+  if [ ++deployed_count == language_count ]
```
jgeewax: Not sure if this is necessary. I think `deploy_docs` will run synchronously, right?

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ev
+
+if [ "${TRAVIS_BRANCH}" != "master" ] || [ "${TRAVIS_PULL_REQUEST}" != "false" ]
```
jgeewax: Maybe add a comment here (I know that I tend to forget bash commands).

This logic says (I think...?):

```
# If we're not on master or it's a pull request, don't do anything:
# Or (flipped around)
# This script only applies for non-pull-requests on master
```

Any chance we can use `== "true"` instead of `!= "false"` ?

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+  then exit 0
+fi
+
+language_count=2
```
jgeewax: I think we don't need these, but I could be wrong.

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+deployed_count=0
+
+function deploy_docs {
+  git submodule add -f -b gh-pages https://${GH_OAUTH_TOKEN}@github.com/$1 gh-pages
+
+  if [ -d gh-pages/src ]
```
jgeewax:
```
# If there is a `gh-pages/src` directory, remove it (we're going to replace it).
```

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+    then git rm -rf gh-pages/src
+  fi
+
+  cp -R src gh-pages
+  cd gh-pages
+  git add src
+  git config user.name "selfiebot"
```
jgeewax: Who owns selfiebot? Maybe we can have googlebot do this? Since it's already checking our CLA signatures?

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+  git commit -m "Update docs for ${TRAVIS_TAG}"
```
jgeewax: Can we clarify which repo this tag is on? We're effectively taking code from gcloud-common, and shove it into gcloud-python, so I'm not sure if the `TRAVIS_TAG` is one attached to gcloud-common or gcloud-python.

deploy-docs.sh
stephenplusplus: Oh, right. I'm not used to that!

deploy-docs.sh
stephenplusplus: It'll be from gcloud-common, but I just noticed that doesn't work for this use case. We probably want `TRAVIS_COMMIT` instead.

A list of these are at http://docs.travis-ci.com/user/environment-variables/

deploy-docs.sh
stephenplusplus: Maybe, do I need its user name and email?

deploy-docs.sh
stephenplusplus: :+1:

deploy-docs.sh
stephenplusplus: Most simply, I think "This script only applies for pushes to master". I'll try to change it to make more sense.

deploy-docs.sh
jgeewax: Sure I'm just saying the comment should be clear that we're pushing from gcloud-common -> gcloud-python.

If the commit says "Update docs for commit d4fde1" in gcloud-python, I'd go looking for that commit in gcloud-python. If it said "Update docs from gcloud-common#d4fde1" (or something like that) it's much more clear.

stephenplusplus: Oh, right. Gotcha.

deploy-docs.sh
stephenplusplus: The != false part is because `TRAVIS_PULL_REQUEST` is either a PR number or the string "false" if it's not a PR. Is there a less weird way to handle this logic?

stephenplusplus:
Updates made!

deploy-docs.sh
```diff
@@ -0,0 +1,46 @@
...
+  # remove it (we're going to replace it).
+  if [ -d gh-pages/src ]
+    then git rm -rf gh-pages/src
+  fi
+
+  cp -R src gh-pages
+  cd gh-pages
+  git add src
+  git config user.name "selfiebot"
+  git commit -m "Pushing a doc site update from https://github.com/GoogleCloudPlatform/gcloud-common/commit/${TRAVIS_COMMIT}"
```
dhermes: Breaks the 72 char width of commit subject.

stephenplusplus: Give it a look now.

deploy-docs.sh
```diff
@@ -0,0 +1,48 @@
...
+    then git rm -rf gh-pages/src
+  fi
+
+  cp -R src gh-pages
+  cd gh-pages
+  git add src
+  git config user.name "selfiebot"
+  git commit -m "Pushing a doc site update
+
+  https://github.com/GoogleCloudPlatform/gcloud-common/commit/${TRAVIS_COMMIT}"
```
dhermes: :+1: (I wasn't not sure the multi-line string will do what you expect, but checked it ang verified it works.)

deploy-docs.sh
```diff
@@ -0,0 +1,48 @@
...
+  then exit 0
+fi
+
+function deploy_docs {
+  git submodule add -f -b gh-pages https://${GH_OAUTH_TOKEN}@github.com/$1 gh-pages
+
+  # If there is an existing `src` directory in the destination's gh-pages branch,
+  # remove it (we're going to replace it).
+  if [ -d gh-pages/src ]
+    then git rm -rf gh-pages/src
```
callmehiphop: Assuming we're building the app via `gulp build`, then there wouldn't end up being a src directory. I think I can tweak to the gulp file to put it there maybe..?

deploy-docs.sh
```diff
@@ -0,0 +1,48 @@
...
+  fi
+
+  cp -R src gh-pages
```
callmehiphop: I think here we'd want to do something like

```
gulp build
cp -R dist/* gh-pages
```

deploy-docs.sh
stephenplusplus: Is there another command we can run to effectively:

`remove docs version 1 cleanly from gcloud-ruby`
`insert docs version 2`?

callmehiphop: yeah, it would essentially be something like 

```
git rm -rf gh-pages/scripts
git rm -rf gh-pages/styles
git rm -rf gh-pages/assets
```

stephenplusplus: Maybe it's better to push it all into a `src` folder then, just thinking if we someday have root level files or another directory, etc.

deploy-docs.sh
stephenplusplus: Thanks! Updated.

deploy-docs.sh
stephenplusplus: What should I do!?

stephenplusplus:
Just calling out this thread, since it got swallowed by an "outdated diff":

https://github.com/GoogleCloudPlatform/gcloud-common/pull/35#discussion_r51448901

deploy-docs.sh
```diff
@@ -0,0 +1,49 @@
...
+  git add src
+  git config user.name "selfiebot"
+  git commit -m "Pushing a doc site update
+
+  https://github.com/GoogleCloudPlatform/gcloud-common/commit/${TRAVIS_COMMIT}"
+  git push https://${GH_OAUTH_TOKEN}@github.com/$1 HEAD:gh-pages
+}
+
+# deploy_docs "googlecloudplatform/gcloud-node"
+# deploy_docs "googlecloudplatform/gcloud-php"
```
callmehiphop: We might want to take PHP off this list, I'm not sure they're ready yet.

dwsupplee: Not quite ready, but it is being worked on currently. Should be ready soon.

deploy-docs.sh
```diff
@@ -0,0 +1,49 @@
...
+  # If there is an existing `src` directory in the destination's gh-pages
+  # branch, remove it (we're going to replace it).
+  if [ -d gh-pages/src ]
+    then git rm -rf gh-pages/src
+  fi
+
+  gulp build
+  cp -R dist/* gh-pages
+  cd gh-pages
+  git add src
```
callmehiphop: This line won't work (there is no `src` directory anymore). Is it naive to suggest `git add --all`?

stephenplusplus: I think it could run into problems. RE: https://github.com/GoogleCloudPlatform/gcloud-common/pull/35#discussion_r51452135 -- is it possible to keep separate the source code of the site coming from this repo and the rest of the gh-pages branch?

callmehiphop: Ahh, ok. Let me see if I can tweak the build script to output everything to a src directory then.

callmehiphop: Ok, so I've made a few alterations to the build and now all css/js/images would be output to a `src` directory. The index.html file would have to stay in the root directory (obviously) which means we'd have to add `git add index.html` as well.

stephenplusplus:
@callmehiphop PTAL!

callmehiphop:
LGTM! :+1:

quartzmo:
Right on!

PR #60, opened 2016-02-02
docs build: set Travis to run Node 0.12

stephenplusplus:
Also, make `npm install` output louder logs.


callmehiphop:
This is it, I can feel it!

PR #65, opened 2016-02-03
site: add ruby support

quartzmo:
Current status: 

[Demo](http://quartzmo.github.io/gcloud-ruby/#/)

[Source](https://github.com/quartzmo/gcloud-ruby/tree/gh-pages)

This PR is to add all that is needed for the site to support Ruby docs. 

So far I've just replaced `node` with `ruby` in `manifest.json`, and therefore needed to make the change that appears in this first commit. @callmehiphop Can you think of any other additions I will need to make?

I'll remove `don't merge` label when everything is working for me locally with our gcloud-ruby json. Creating the PR now to get help and feedback.

TODO:

- [x] Service description content should render markdown code blocks
- [x] Ruby code highlighting
- [x] Fix links to images
- [x] Convert old rdoc examples to YARD
- [x] Add "see also" links to method descriptions
- [x] Add "see also" links to class descriptions
- [x] Make sure all classes have titles/headings in the description (@blowmage to do later)
- [x] Make sure all params (including yield) have a description
- [x] Link to old (external) docs from version switcher, since we can't produce JSON for old versions (Extracted to #90)
- [x] Fix internal links


quartzmo:
Btw, I don't see any tests to run, but I did test the change manually with `gulp serve` and ran `gulp build` locally. (Using Node 0.12).

callmehiphop:
You're going to want to add a [hook for mapping custom types](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/components/custom-type/custom-type.directive.js#L11).  We ended up going with @blowmage's suggestion of creating `<a>` tags with a data attribute - `data-custom-type`. 

> I don't see any tests to run

We didn't have any pre-existing tests with the old site, so I never made any, but I believe it would be wise to invest in some. @jgeewax thoughts on that + timing?

quartzmo:
I spent the day changing our JSON generator to produce multiple files, and it works! Added 
[Demo](http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/gcloud) and [Source](https://github.com/quartzmo/gcloud-ruby/tree/gh-pages) links to the description above.

callmehiphop:
Awesome! Looks good so far :+1:

quartzmo:
@callmehiphop Take a look now, @blowmage helped to push some serious progress today! We now have all of our code objects published, including those that are nested a level deep within the service objects.

callmehiphop:
@quartzmo overall I think it's looking pretty good. Here are a few things I'm noticing

* The logo has an external link icon, did the url get changed?
* Nested items are denoted as `BigQuery -> Dataset::List`, should we update the side navigation to allow for nested sections to make it appear as `BigQuery -> Dataset -> List`?
* Some of your overviews have code examples that don't have syntax highlighting -(http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/gcloud/backoff). I didn't anticipate code samples to be in there, do you think it would make more sense to use markdown here?
* Still missing custom type linking (http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/copyjob) - see above comment
* You guys don't have an overview, I just sent a PR to optionally display the "Getting started with gcloud" section. - https://github.com/GoogleCloudPlatform/gcloud-common/pull/72
* Haven't gone through all the params, but I see a few missing descriptions (http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/dataset?method=access)
* Looks like some of the URLs aren't coming through in the "More Information" section (http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/dataset?method=access)
* It also looks like some pages you add a heading to the description (http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/dataset-access) and some of them you don't (http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/dataset-list).

I think we're in pretty good shape. Is there anything I can to help? Missing features on the site that would make your lives easier, etc.?

blowmage:
Thanks for the feedback! Here are some thoughts.

> * The logo has an external link icon, did the url get changed?

Not sure about that. I'll add it to the list of things to follow up on.

> * Nested items are denoted as BigQuery -> Dataset::List, should we update the side navigation to allow for nested sections to make it appear as BigQuery -> Dataset -> List?

We discussed and investigated it. This would require more changes than we are comfortable making until we get closer to feature complete/compatible. We want to suggest smaller changes instead of large ones. We also have thoughts about how navigation should work. In our current docs we don't show all classes/modules in the side nav, only the most important ones. (As requested by JJ.) But we cannot do that with the angular code as is. There is no way to have a class/module that is not also listed in the side nav. So for now we are comfortable listing everything, but we will follow up with some specific suggestions later.

> * Some of your overviews have code examples that don't have syntax highlighting

Yeah, we had to stop last night before we could add [the CSS](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/gh-pages/css/main.css#L1596-L1656) for that. It will be fixed soon. Those code examples are written using markdown, and the generated markup is pygments compatible.

> * Still missing custom type linking

We are actively working on it. There are assumptions made in the angular app's URL routing and JSON data structure that make those anchor tags difficult to generate and resolve for us, and we assume difficult for the other languages. We have discussed possible alternative approaches that would make this much easier for us, but we want to get feature complete/compatible before suggesting them.

> * You guys don't have an overview

Added to the list.

> * Haven't gone through all the params, but I see a few missing descriptions

Added to the list.

> * Looks like some of the URLs aren't coming through in the "More Information" section 

Added to the list.

> * It also looks like some pages you add a heading to the description

Added to the list.

callmehiphop:
> We also have thoughts about how navigation should work

Might be relevant: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1084

> There is no way to have a class/module that is not also listed in the side nav.

Gotcha, we could probably remedy that pretty easily by adding a new key to the `manifest.json` that stated whether or not it should show up in the side nav.

> There are assumptions made in the angular app's URL routing and JSON data structure that make those anchor tags difficult to generate and resolve for us, and we assume difficult for the other languages.

What if we created an `href` key within the services/guides manifest objects? Instead of having dynamic routes, we could just let the manifest dictate that.

blowmage:
@callmehiphop I feel like communicating our ideas over GitHub issues is not going to be as productive as a hangout or even slack chat. Would that be possible?

In a nutshell, here are the tweaks we'd like to see:

1) Separate the manifest into a main JSON that lists versions, and multiple version JSON files that contain the navigation for that version.
2) Separate the navigation structure from the list of code objects in the version JSON. The code objects can point to an object JSON that has all the documentation. This means that we can have code objects that are not in the navigation, and will enable new features like search.
3) Give every guide, code object, and code method a unique id. The id is the value of the `data-custom-type` in the anchor tags, and it what is put into the URL, making resolving it much easier.

I have some sample JSON if this structure if you'd like to see it.

callmehiphop:
Hangouts is cool with me! And sure, maybe throw it in a gist?

quartzmo:
@callmehiphop Is the [`home.html` in gcloud-node gh-pages](https://github.com/GoogleCloudPlatform/gcloud-node/blob/gh-pages/home.html) generated? Or copied from somewhere? The reason that image links are broken in our `home.html` is that they use `assets/images/` path (good for development, but not for production, which needs `src/images/`, as in your gh-pages home.html). How do you manage the path difference for production?

callmehiphop:
@quartzmo <kbd>home.html</kbd> is not generated, so you have to manually update the paths. :disappointed: 

It seemed like [the consensus](https://github.com/GoogleCloudPlatform/gcloud-common/issues/44) was not to use JSON to create that screen, and in an effort to remove old (and possibly unused) site files easily, we [recently changed](https://github.com/GoogleCloudPlatform/gcloud-common/pull/56) the directory where all that stuff lived. I kinda forgot about that file, hence why the paths are funky.

quartzmo:
@callmehiphop OK, no problem at all, just thought I'd check with you. Converting our old home file now.

quartzmo:
@callmehiphop Can we add a list of `resource.link` & `resource.title` pairs for classes as well as methods? Currently I only see this in the [method section](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/service/service.html#L71-L81). [Our JSON](https://github.com/quartzmo/gcloud-ruby/blob/gh-pages/json/master/bigquery/table.json) includes this:

```js
{
  id: "table",
  metadata: {
    name: "Table",
    description: "<h1 id="table">Table</h1> <p>A named resource representing a BigQuery table that holds zero or more records. Every table is defined by a schema that may contain nested and repeated fields.</p>",
    source: "lib/gcloud/bigquery/table.rb#L66",
    resources: [
      {
        title: "Preparing Data for BigQuery",
        link: "https://cloud.google.com/bigquery/preparing-data-for-bigquery"
      }
    ],
    ...
```

I would be happy to open a new issue and prepare a PR for this.

callmehiphop:
When you say class section, are we referring to the constructor or the actual class overview section? If it's the constructor, the resources should just work. If it's the class overview, we would need to add that, but I think it would be a pretty quick change!

quartzmo:
@callmehiphop: class overview

Definitely a quick addition, but since we have a lot of links like this in our docs, a big win! My 2 cents is to put this section [just after the description](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/service/service.html#L22) and before the method listings. For me, it's really a part of the description. If you put it after the methods, readers interested in "the big picture" probably won't get the info.

callmehiphop:
:+1: I can dig it, would you be interested in making a PR for this?

quartzmo:
@callmehiphop Yes, happy to put together a PR.

quartzmo:
@callmehiphop Any ideas about how to do this one?

> Link to old (external) docs from version switcher, since we can't produce JSON for old versions

I guess we need to expand the manifest JSON to allow some versions to map to external URLs? And add logic in `DocsCtrl#loadVersion` to handle such versions with redirection the mapped URL? Awkward, right? Any better idea? Should I extract this to a new issue?

callmehiphop:
@quartzmo I think a new issue is a good idea, gcloud-python may be in a similar boat.

quartzmo:
@callmehiphop @blowmage I believe this PR completes basic Ruby support and is good to merge. My final commit adds a mapping for Ruby to the same custom type handling used for Node. It also adds logic in the `customType` directive for using the element's existing text, if present, to allow alternate link text. For example:

```html
<a data-custom-type="bigquery/dataset">Gcloud::Bigquery::Dataset</a>
```

To see the updated `customType` directive in action, check out [the overview for Bigquery::Project on my gh-pages](http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery/project).

@callmehiphop I noticed that  the `customType` directive does not pass library version when linking to a new route, so that if you're not on the default version, you end up getting switched to it. Should I open an issue for that?

callmehiphop:
@quartzmo nice! I haven't spent a lot of time this week moving to @blowmage's schema updates, but I guess my question for you both is whether or not you'd prefer to hold off on merging until that's addressed since the current format leaves certain aspects of your docs in an undesirable state.

quartzmo:
@callmehiphop I am working incrementally, so I am in favor of merging this PR now, which will allow us to get back to the `master` version of the site app for our gh-pages demo. Also, this PR does not change much code; but significantly, it adds some Ruby-specific images that @blowmage lovingly created.

callmehiphop:
@quartzmo awesome! If it's alright with you and @blowmage I'd like to wait until tomorrow to merge this in - I'm going to see Deadpool in a bit and if anything went South I'd hate to leave you guys hanging.

blowmage:
So jelly. I have to wait until tomorrow to see it...

quartzmo:
@blowmage Nah, you can see it now! [Demo](http://quartzmo.github.io/gcloud-ruby/#/)

jgeewax:
That is so beautiful. You win. :+1: 

Question for you: multiple examples don't seem to have the same spacing when I compare between Ruby / Node. For example:

http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/dns/zone

vs

http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.28.0/pubsub?method=getSubscriptions

Do we have CSS differences floating around ?

quartzmo:
@jgeewax The CSS should be identical except for Ruby syntax highlighting. Did you mean the spacing between examples?

Ruby:

ruby-spacing

Node:

node-spacing

If so, this is due to differences in HTML.

Ruby:

```html
<div ng-if="example.caption" bind-html-compile="example.caption" class="ng-scope">
  <span class="ng-scope">The changes can be sorted by change sequence:</span>
</div>
```

Node:

```html
<div ng-if="example.caption" bind-html-compile="example.caption" class="ng-scope">
  <p class="ng-scope">To control how many API requests are made and page through the results manually, set <code>autoPaginate</code> to <code>false</code>.</p>
</div>
```

In this case, the Node caption is enclosed in a `p` element. This comes from the JSON. ([Ruby](https://raw.githubusercontent.com/quartzmo/gcloud-ruby/gh-pages/json/v0.6.1/dns/zone.json) vs [Node](https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-node/gh-pages/json/v0.28.0/pubsub/index.json))

Do you see any other spacing differences?

quartzmo:
Regarding a fix, it wouldn't be hard to wrap the Ruby captions in `p` element when we generate JSON, but it does seem like a slippery slope to depend on markup with the JSON. In our case, YARD does not allow us to have multi-paragraph captions, just a single line per example.

jgeewax:
Ah gotcha. Maybe we can pull that out of the JSON (from gcloud-node) and move it over to the Angular app? (And then Ruby will just get it for free?)

Obviously it's a subtle difference, so not the end of the world, but what I'm saying is....

- I think the spacing as it appears on Node's docs is prettier.... :)
- I think the fact that Node gets that by putting a `<p>` tag in it's JSON is coupling content and style a bit too tightly :(

Am I crazy?

callmehiphop:
@quartzmo I noticed that some of the custom type linking didn't appear to be working, but I realized that each one was actually for some of the hidden pages (e.g. *Gcloud::Dns::Change*). Is part of our plan to add support for these later?

I also am curious about the markdown css for ruby, we currently use a library called [highlightjs](https://highlightjs.org/) that has Ruby syntax support, if possible I think we should try and leverage this. Can I ask which examples (guides, overviews, etc.) needed this additional stylesheet?

callmehiphop:
> Ah gotcha. Maybe we can pull that out of the JSON (from gcloud-node) and move it over to the Angular app? (And then Ruby will just get it for free?)

Totally doable!

> I think the fact that Node gets that by putting a <p> tag in it's JSON is coupling content and style a bit too tightly 

Aside from the example captions, this is currently unavoidable for us, dox doubles up as a markdown and JSDoc parser, so it will inject html into most of our fields (descriptions, etc.).

callmehiphop:
Probably worth mentioning that the tradeoff of Node moving from paragraphs to spans is that we (obviously) will have to contain ALL text within a single paragraph.. so anywhere we had multiple paragraphs explaining a single example will be gone.

quartzmo:
@callmehiphop That's a great point. Also, I just looked and I actually have some post-processing in our JSON generator that removes the `p` elements that our markdown lib produces by default! 

```ruby
html = unwrap_paragraph(html) unless multi_paragraph
...

def unwrap_paragraph html
   match = Regexp.new(/\A<p>(.*)<\/p>\Z/m).match(html)
   match[1] if match
end
```

If i disable this operation, my tests fail as follows:

```sh
  6) Failure:
Gcloud::Jsondoc::module::when a module has a method#test_0002_must have metadata examples [/Users/quartzmo/code/google/codez/gcloud-jsondoc/test/module_test.rb:49]:
Expected: "You can pass options."
Actual: "<p>You can pass options.</p>"
```

So, I think we can safely say "wrapping in a p element is just what markdown does" (although this is an empirical conclusion for me), and change the Ruby JSON.

quartzmo:
@callmehiphop Regarding the code highlighting CSS, @blowmage and I just followed the overwhelming favorite of our community (and our markdown lib) and took the path of much less resistence with [Pygments](http://pygments.org/). Just to give you an idea of Pygments' dominance, compare 2 *million* downloads for [pygments.rb](https://rubygems.org/gems/pygments.rb/) with 4 *thousand* downloads for [hljs (ruby)](https://rubygems.org/gems/hljs).

We could try to get hljs working with our markdown lib, or attempt to do the highlighting post-markdown, but if it's ok with you can we wait and see if other languages also want/need Pygments styling? For Python it seems like sure thing.

quartzmo:
@callmehiphop Regarding:

> It also looks like maybe there's a bug with the custom-type linking.. if you look at #update links at http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/dns/zone?method=import

I should have mentioned before that I have been unable to figure out why YARD (our docs lib) is failing to resolve these instance methods, even when I give it the correct code object context containing them. I'll keep working on it. The solution will be in our JSON generation, and should not impact this PR.

quartzmo:
@callmehiphop @jgeewax @blowmage I've just managed to fix the internal method links issue with [gcloud-ruby PR #555](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/555), which also now leaves example captions wrapped in `p` elements (the default markdown output) for the same examples spacing as Node docs.

See [DNS::Zone in the Demo](http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/dns/zone) to experience both of these exciting fixes.

quartzmo:
My goodness, just when I thought there was some light in the tunnel...

Please hold off on merging this while I investigate why our service overview [code examples have no line breaks](http://quartzmo.github.io/gcloud-ruby/#/docs/v0.6.1/bigquery).

quartzmo:
I restored the line breaks to the output.

@callmehiphop I believe this PR is ready to merge.

callmehiphop:
@quartzmo LGTM :)

quartzmo:
Thanks @callmehiphop!

Issue #54, opened 2016-02-01
Add documentation about our documentation build process


stephenplusplus:
I think it would be helpful to have a description of the way docs are built and how each place (gcloud-common + gcloud-\*) plays a role in the process as well as what a gcloud-\* library has to do to be compliant. Maybe not anything extensive (@callmehiphop you've got a lot going on!), but just something that answers some obvious questions.

callmehiphop:
+1

jgeewax:
![we-need-to-go-deeper](https://cloud.githubusercontent.com/assets/112928/12726971/8c73b5a2-c8e8-11e5-8d4b-3597a51f9f5a.jpg)

dhermes:
:+1: on the meme

stephenplusplus:
![onthememe](https://cloud.githubusercontent.com/assets/723048/12727253/1c4e6158-c8ea-11e5-87dc-627447778074.png)

jgeewax:
And @stephenplusplus wins.

PR #67, opened 2016-02-03
docs docs: I heard you like docs

callmehiphop:
Closes https://github.com/GoogleCloudPlatform/gcloud-common/issues/54

I've made up a basic table of contents for the areas that (I think) could use some documentation. Let me know if you think I'm missing some important features or if something doesn't make sense, etc.

[demo!](https://github.com/callmehiphop/gcloud-common/blob/docs-inception/site/README.md)

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -1,0 +1,495 @@
+# gcloud Shared Docs App
+
+## Table of Contents
+
+* [Installation][installation]
+* [Previewing Locally][previewing]
+  * [Previewing with `gh-pages`][previewing-ghpages]
+* [Manifest][manifest]
+  * [`lang`][lang-key]
+  * [`friendlyLang`][friendlylang-key]
+  * [`markdown`][markdown-key]
+  * [`versions`][versions-key]
+  * [`content`][content-key]
+  * [`home`][home-key]
+  * [`overview`][overview-key]
+  * [`guides`][guides-key]
+  * [`services`][services-key]
+  * [`package`][package-key]
+* [JSON Docs Schema][json-schema]
+  * [`id`][id-key]
+  * [`description` and `caption`][description-keys]
+  * [`metadata`][metadata-key]
+  * [`methods`][methods-key]
+    * [`metadata`][metadata-key-1]
+    * [`params`][params-key]
+    * [`exceptions`][exceptions-key]
+    * [`returns`][returns-key]
+  * [Custom Data Types][custom-types]
+* [Misc. Content][misc-content]
+  * [Landing Page][landing-page]
+  * [Overview Section][overview-section]
+  * [Custom Page Headers][custom-page-headers]
+* [Deploying][deploying]
+
+## Installation
+
+Install `bower` and `gulp` globally. You'll need to install [Node.js][nodejs] if it isn't already.
+
+```sh
+$ npm install -g gulp-cli bower
+```
+
+Clone the gcloud-common repository
+
+```sh
+$ git clone git@github.com:GoogleCloudPlatform/gcloud-common.git
+$ cd gcloud-common/site
+```
+
+Install all the things!
+
+```sh
+$ bower install
+$ npm install
+```
+
+Now we can start the local webserver via `gulp`. This will launch a browser window.
+
+```sh
+$ gulp serve
+```
+
+## Previewing Locally
+
+In order to test locally you'll need a [<kbd>manifest.json</kbd>][manifest] file as well as content directory containing your [JSON docs][json-schema].
+
+For testing purposes these should both be placed directly within the `src` directory.
+
+Your <kbd>src</kbd> directory should then resemble something like the following
+
+```sh
+gcloud-common/site/src
+ ├─── app # front-end application files
+ ├─── assets # images, etc.
+ ├─── versions # JSON content folder
+ |     └─── v0.28.0 # one folder per release
+ |           └─── storage # one folder per service
+ |                 ├─── index.json # one file per class/module/etc.
+ |                 └─── bucket.json
+ ├─── index.html
+ └─── manifest.json
+```
+
+For a living example, refer to gcloud-node's [gh-pages][gcloud-node-ghpages]. *Please note* their content directory is called <kbd>json</kbd>.
+
+### Previewing with `gh-pages`
+
+If you already have the docs app deployed to your gcloud library's `gh-pages` branch, this can make testing a whole lot easier. Since the application is already built, you can simply checkout the `gh-pages` branch and add your JSON to the <kbd>master</kbd> directory.
+
+**Note:** some browsers may require you to disable web security to make local http requests, you can circumvent this by running a local web server.
+
+## Manifest
+
+The <kbd>manifest.json</kbd> is used to supply the Angular application with the necessary information to display your documentation properly. Please refer to the [schema][manifest-schema] for more information.
+
+##### `lang` key
+
+This is the non-friendly name for your language, it will be used to generate links to github, stackoverflow, etc. It's also used internally within the application for [language specific hooks][custom-page-headers]
+
+```js
+{
+  "lang": "node"
+}
+```
+
+##### `friendlyLang` key
+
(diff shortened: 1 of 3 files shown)


jgeewax:
compelling

site/README.md
```diff
@@ -0,0 +1,47 @@
+# gcloud Shared Docs App
+
+## Table of Contents
+
+* [Installation](#installation)
```
jgeewax: Can't we get this for free with the .wiki format ? Bums me out that we have to do this ourselves...

callmehiphop: I can definitely move this to gcloud-common wiki once a solid draft has been written, I just like PRs for the raised visibility and feedback format.

stephenplusplus:
:eyes: :+1: :v: :man:

callmehiphop:
> :eyes: :+1: :v: :man:

I almost read that as "looks good to me", but then I remembered you don't have a mustache.

stephenplusplus:
Looks good to Freddie Mercury.

callmehiphop:
I've updated the description with a demo of the README.md.

PTAL!

callmehiphop:
@quartzmo @blowmage since you've already gotten up an running with the docs app locally, could you take a look through this and let me know if you think anything is missing/wrong/etc.?

@dwsupplee not sure how far along you are with any of this, but if you wouldn't mind taking a run through, that would be awesome.

dwsupplee:
You got it @callmehiphop. I'll be running through it tonight and tomorrow.

quartzmo:
@callmehiphop Same here, will take a look!

callmehiphop:
@dwsupplee @quartzmo thanks!

site/README.md
```diff
@@ -0,0 +1,483 @@
...
+$ npm install
+```
+
+Now we can start the local webserver via `gulp`. This will launch a browser window.
+
+```sh
+$ gulp serve
+```
+
+### Testing Locally
```
quartzmo: I expected this section (and the following one) to be about automated testing. Maybe title with "Preview" instead?

site/README.md
```diff
@@ -0,0 +1,483 @@
...
+ |     └─── v0.28.0 # one folder per release
+ |           └─── storage # one folder per service
+ |                 ├─── index.json # one file per class/module/etc.
+ |                 └─── bucket.json
+ ├─── index.html
+ └─── manifest.json
+```
+
+For a living example, refer to gcloud-node's [gh-pages][gcloud-node-ghpages]. *Please note* their content directory is called <kbd>json</kbd>.
+
```
quartzmo: I'm so confused!™ Why the flexibility on the directory name? Can we stick with one or the other?

site/README.md
```diff
@@ -0,0 +1,483 @@
...
+For a complete example of a service JSON, please refer to [this gist][json-schema-gist].
+
+##### `id` key
+
+This is the id for the service.
+
+```js
+{
+  "id": "bucket"
```
quartzmo: I think it would be easier to follow these key examples if they maintained continuity with the first (BigQuery) example shown. In other words, stick to the BigQuery example throughout.

quartzmo:
@callmehiphop One area that has been really fuzzy for @blowmage and me, and which I would love to see detailed, is exactly what HTML elements are allowed/expected in `description`, `caption`, etc. For example, we have completely formatted and styled code examples in our descriptions. And remember how captions *without* `p` wrappers did not display correctly?

site/README.md
callmehiphop: Good call!

site/README.md
callmehiphop: I just wanted to provide maximum flexibility. I think you're right though, it does seem unnecessary. Sorry for the confusion!

site/README.md
callmehiphop: :+1:

callmehiphop:
@quartzmo thank you for all the feedback! I went through and made some adjustments.

> exactly what HTML elements are allowed/expected in description, caption, etc.

From what I can tell, `dox` will wrap descriptions in either a `<p>` or `<ul><li></li></ul>`. Not sure what's up with the unordered list, but essentially either will work because of the browsers default margin/padding  (we don't cancel those CSS rules out). That being said, just about any HTML element *will* work, so long as it has default padding and/or margin.

This is obviously pretty undesirable and I would be open to creating a new issue to removing the HTML from our JSON and adding it to the application itself.

site/README.md
```diff
@@ -0,0 +1,495 @@
...
+Different languages may wish for different formatting when it comes to page titles, when moving to the common-docs you can design a [template][page-header-template] to display the title how you wish.
+
+A `title` array containing the name hierarchy will be available for you.
+
+When visiting `/docs/latest/storage` the `title` array will resemble `['Storage']`.
+When visiting `/docs/latest/storage/bucket` the `title` array will resemble `['Storage', 'Bucket']`.
+
+## Deploying
+
+Integrating the shared docs app is fairly painless. There is a [shell script][gcloud-common-deployhttps://github.com/GoogleCloudPlatform/gcloud-common/blob/master/deploy-docs.sh#L71] that will push the updated site code to your `gh-pages` after a successful merge. To enable this, simply append a command to push to your repo to the script.
```
stephenplusplus: This [shell-script] link is busted.

site/README.md
```diff
@@ -0,0 +1,495 @@
...
+
+See a living example of gcloud-node's home template [here][gcloud-node-home].
+
+##### Overview Section
+
+Specify an overview using the <kbd>manifest.json</kbd>'s [`overview` key][overview-key].
+
+Overviews are completely optional, you can create one using html.
+
+See [gcloud-node docs][gcloud-nod-docs] for a working demo. The contents of the overview file fill the *Getting Started with gcloud* section of the service pages.
```
stephenplusplus:
```diff
-[gcloud-nod-docs]
+[gcloud-nod-docs]
```

stephenplusplus:
The only thing I see is maybe bringing the JSON schema files into this repo instead of gist files might be helpful. It would also track progression as we change it.

callmehiphop:
> The only thing I see is maybe bringing the JSON schema files into this repo instead of gist files might be helpful. It would also track progression as we change it.

:+1:

quartzmo:
@callmehiphop I think in the short term, just regarding this PR, updating the examples to contain the expected HTML markup is the way to go.

site/README.md
```diff
@@ -0,0 +1,495 @@
...
+      "description": "<p>The callback function.</p>",
+      "types": ["function"],
+      "optional": true,
+      "nullable": false
+    }]
+  }]
+}
+```
+
+It's also possible to document nested keys or callback arguments by simply prefixing the key name with the object/callback name.
```
dwsupplee: A little more detail on how many levels deep can you nest would be very helpful here.

dwsupplee:
This is a huge help in getting the docs site up and running for PHP. Awesome work @callmehiphop!

Have you considered adding a link out to a [json-schema](http://json-schema.org/examples.html) for the JSON documents? It could help provide some additional clarity for which fields are required and allow us to validate our JSON as part of the build process.

callmehiphop:
> Have you considered adding a link out to a json-schema for the JSON documents?

@stephenplusplus suggested the same thing, I'm actually in the progress of doing just that!

site/README.md
callmehiphop: Technically there is no limit, check out the [gcloud parameters](https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.27.0/gcloud) for an example of how the nested properties are displayed.

callmehiphop:
I've checked in some schema files for `manifest.json` and `service.json`. PTAL!

blowmage:
:+1: for using JSON schemas. I think they will help tremendously. I also included them in my [gist](https://gist.github.com/blowmage/13be352f40ee7ac5ee48) for some suggested changes.

dwsupplee:
Looks great!

callmehiphop:
This has been kind of idle, if there's nothing else I'll just merge?

quartzmo:
@callmehiphop SGTM

PR #71, opened 2016-02-04
site: catch-all route logic

callmehiphop:
Relates to: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1112

Adding logic to redirect user from unknown routes.

### DIFF
--- a/site/src/app/docs/docs.route.js
+++ b/site/src/app/docs/docs.route.js
@@ -7,6 +7,10 @@
 
   /** @ngInject */
   function docsRoutes($stateProvider, $urlRouterProvider, $urlMatcherFactoryProvider, manifest) {
+    // source: https://github.com/sindresorhus/semver-regex
+    var regSemver = '\\bv?(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)(?:-[\\da-z\-]+(?:\\.[\\da-z\\-]+)*)?(?:\\+[\\da-z\\-]+(?:\\.[\\da-z\\-]+)*)?\\b';
+    var latestVersion = manifest.versions[0];
+
     $urlMatcherFactoryProvider.type('nonURIEncoded', {
       encode: toString,
       decode: toString,
@@ -15,7 +19,14 @@
 
     $stateProvider
       .state('docs', {
-        url: '/docs/:version',
+        // ui-router allows one to optionally use a regular expressions to
+        // match uri parameters against. In this instance we're only allowing
+        // the version value to be "master" OR a valid semver version.
+        // If we receive a route that does NOT match one of these two values
+        // then we'll end up in the $urlRouterProvier.otherwise(...) block
+        // which will look for version alias's or the absense of a version
+        // and redirect the user appropriately
+        url: '/docs/{version:master|' + regSemver + '}',
         templateUrl: 'app/docs/docs.html',
         controller: 'DocsCtrl',
         controllerAs: 'docs',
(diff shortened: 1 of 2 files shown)


callmehiphop:
@stephenplusplus PTAL

site/src/app/docs/docs.route.js
```diff
@@ -15,7 +17,7 @@
 
     $stateProvider
       .state('docs', {
-        url: '/docs/:version',
+        url: '/docs/{version:master|v\\d+\.\\d+\.\\d+}',
```
stephenplusplus: Can you leave a code comment about wat this is.

site/src/app/docs/docs.route.js
```diff
@@ -46,6 +48,28 @@
...
+      var params = $location.path().replace(baseUrl, '').split('/');
+      var isValidVersion = versions.indexOf(params[0]) !== -1;
+
+      // could be a bad service name
+      if (isValidVersion) {
+        return baseUrl + params[0];
+      }
+
+      // could be a version alias
+      if (params[0] === 'latest' || params[0] === 'stable') {
```
stephenplusplus: Are any places linking to `/stable` or `/latest` already?

callmehiphop: Not sure, I've seen this used in both Ruby and Python though, so I decided to include it.d

stephenplusplus: That's cool. I was just thinking if we don't use it already, we could ditch it. But not breaking existing URLs is a way better option!

site/src/app/docs/docs.route.js
callmehiphop: Done!

site/src/app/docs/docs.route.js
```diff
@@ -15,7 +17,14 @@
...
       .state('docs', {
-        url: '/docs/:version',
+        // ui-router allows one to optionally use a regular expressions to
+        // match uri parameters against. In this instance we're only allowing
+        // the version value to be "master" OR a valid semver version.
+        // If we receive a route that does NOT match one of these two values
+        // then we'll end up in the $urlRouterProvier.otherwise(...) block
+        // which will look for version alias's or the absense of a version
+        // and redirect the user appropriately
+        url: '/docs/{version:master|v\\d+\.\\d+\.\\d+}',
```
stephenplusplus: I love that comment. Thanks, sir!

But on the semver regex: https://github.com/sindresorhus/semver-regex/blob/master/index.js

It's not always just digits. Should we just borrow that regex and link to it in a comment?

callmehiphop: We could do that, do you think we'll end up including releases on the doc site that include a suffix? (e.g. x.x.x-pre)

stephenplusplus: It's possible, since we support everything from `master` to `stable`. I just see it as an unexpected, not obvious to find bug that could smack us in the future.

callmehiphop: Ok, I've added the regex, PTAL!

site/src/app/index.run.js
```diff
@@ -16,10 +16,15 @@
...
 
-    // uncomment for debugging
-    // $rootScope.$on('$stateChangeError', function() {
-    //   console.log(arguments);
-    // });
+    $rootScope.$on('$stateChangeError', function() {
+      // uncomment for debugging
+      // console.log(arguments);
+
+      $state.go('docs.service', {
```
stephenplusplus: Should we add a @todo for whipping up a 404 page?

stephenplusplus: Or I guess the modern version of that, just an issue on GitHub, will work.

callmehiphop: Good idea, new issue?

PR #72, opened 2016-02-05
site: add check for overview file

callmehiphop:
Inserted a conditional to check if we should be displaying the library overview/quick start guide on service pages.

### DIFF
--- a/site/src/app/docs/docs.controller.js
+++ b/site/src/app/docs/docs.controller.js
@@ -16,15 +16,20 @@
     docs.services = angular.copy(manifest.services).filter(isAvailable);
     docs.services.forEach(updateNav);
     docs.version = $state.params.version;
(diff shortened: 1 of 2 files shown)


site/src/app/docs/docs.controller.js
```diff
@@ -16,16 +16,21 @@
...
-      manifest.content,
-      $state.params.version,
-      manifest.overview
-    ].join('/');
     docs.loadVersion = loadVersion;
     docs.getGuideUrl = getGuideUrl;
     docs.isActive = isActive;
 
+    if (manifest.overview) {
+      docs.overview = [
```
stephenplusplus: Can this be a more literal name, like `overviewFileUrl`?

callmehiphop: sure!

PR #97, opened 2016-02-18
Support any JSON path in docsRoutes

### LINKED ISSUE #78: Separate Code Data from Navigation
The current services data structure performs double duty. First, it defines the navigation structure, and second it defines the code constructs that are to be documented. These are two different things, and we would like to suggest that they be separated and replaced with two different data structures.

In gcloud-ruby we have many more code constructs (classes, modules) than gcloud-node. Because of this we do not list all of the code constructs in the navigation, prefering to show only the most important constructs. Unfortunately the current angular app is unable to render a code construct's information unless it is also listed in the navigation. This is a problem for us.

Another reason to split these two structures is that for gcloud-ruby's case, and we suspect for other languages, the navigation items listed does not match the namespace structure of our code. We would prefer a structure that more closely resembles the code constructs, and a separate structure to define how the code constructs are to be navigated.

quartzmo:
[refs #78]

This change supports existing routing, titles, and JSON, while adding support for:

* Optional `metadata.title` in content JSON file
* Nesting of content JSON files to any depth
* Navigation to content JSON data that is not listed in a service `nav`

I am not necessarily proposing that this change closes #78, since @blowmage may want to continue work on that issue to enable search and other goals. However, despite being just an incremental (and backwards-compatible) change, it does remove gcloud-ruby's [no. 1 blocker](https://github.com/GoogleCloudPlatform/gcloud-common/issues/78#issuecomment-184798625).

### DIFF
--- a/site/src/app/docs/docs.route.js
+++ b/site/src/app/docs/docs.route.js
@@ -119,25 +119,33 @@
   /** @ngInject */
   function getService($state, $stateParams, $interpolate, $http, manifest, util) {
     var ids = $stateParams.serviceId.split('/');
-    var serviceId = ids[0];
-    var pageId = ids[1];
+    var serviceId = ids.shift();
+    var pageId = ids.join('/');
     var service = util.findWhere(manifest.services, { id: serviceId });
     var pageTitle = service && service.title ? [service.title] : null;
+    var resource = service.contents;
 
     if (service && pageId) {
-      service = util.findWhere(service.nav, { id: pageId });
-      pageTitle.push(service.title);
+      var page = util.findWhere(service.nav, { id: pageId });
+      if (page) {
+        resource = page.contents;
+        pageTitle.push(page.title);
+      } else {
+        resource = $stateParams.serviceId + '.json';
+      }
     }
 
-    var json = $interpolate('{{content}}/{{version}}/{{data}}')({
+    var json = $interpolate('{{content}}/{{version}}/{{resource}}')({
       content: manifest.content,
       version: $stateParams.version,
-      data: service.contents
+      resource: resource
     });
 
     return $http.get(json).then(function(response) {
       var data = response.data;
-      data.metadata.title = pageTitle;
+      if (typeof data.metadata.title === 'undefined') {
+        data.metadata.title = pageTitle;
+      }
       return data;
     });
   }


site/src/app/docs/docs.route.js
```diff
@@ -119,25 +119,33 @@
...
       content: manifest.content,
       version: $stateParams.version,
-      data: service.contents
+      resource: resource
     });
 
     return $http.get(json).then(function(response) {
       var data = response.data;
-      data.metadata.title = pageTitle;
+      if (data.metadata.title === undefined) {
```
callmehiphop: Small nitpick here, can we check if the title is undefined via `typeof`?

callmehiphop:
@quartzmo I had one small nitpick, but otherwise it looks great! I also agree that this shouldn't close #78.

site/src/app/docs/docs.route.js
jgeewax: Do we have a linter anywhere that would make a build fail for this kind of stuff?

quartzmo: @jgeewax Whether there's a linter for gcloud-common site PRs is a good question, and there should be one--but I don't think it would have failed for this comparison. 

Try it for yourself with the console at [jslint.com](http://www.jslint.com/).

Paste in:

```javascript
var a;
if (typeof a === 'undefined') {
    a = 1;
}
```

It fails  with the message: "Unexpected 'typeof'. Use '===' to compare directly with undefined."

Then try:

```javascript
var a;
if (a === undefined) {
    a = 1;
}
```

It passes.

[Explanation here](https://plus.google.com/101248256976407044060/posts/Q5oFnnxG9oL) (Scroll to Ron Walden comment; although Crockford's replies are hilarious.)

quartzmo: @callmehiphop I'm still happy to change it to use `typeof` if you want. I'm sure the less psycho linters will never complain about it, since it's so common.

callmehiphop: @jgeewax we do have a lint task in our gulp script, although I don't believe it's tied to CI, but I'd gladly hook that up.

@quartzmo if you don't mind!

quartzmo: @callmehiphop I don't mind at all. Just working on linking the return types in our JSON, which should not affect this PR, but I want to be certain.

quartzmo:
@callmehiphop Updated with `typeof` check!

callmehiphop:
:+1: awesome!

PR #99, opened 2016-02-22
add PHP support to site

dwsupplee:
This PR adds PHP support to the docs site and includes a few minor changes.

- The 'available methods' section now shows up under the overview section as opposed to the constructor.
- Constructors now show up in the 'available methods' section.
- The constructor is now styled and linkable like the rest of the methods.

[demo](http://dwsupplee.github.io/gcloud-php/#/)

Todo:
- [x] Publish gcloud-php to packagist (Goal: end of q1) - Track progress here: https://github.com/GoogleCloudPlatform/gcloud-php/pull/24
- [x] Add PHP icon
- [x] Fix encoding of examples found in overview.html and home.html
- [x] Remove references to `gcloud` per https://github.com/GoogleCloudPlatform/gcloud-php/issues/22
- [x] Merge https://github.com/GoogleCloudPlatform/gcloud-php/pull/42


callmehiphop:
:+1: 

Could you link us to a demo?

dwsupplee:
Sure, please keep in mind it is still a work in progress.

http://dwsupplee.github.io/gcloud-php/#/

quartzmo:
@dwsupplee Here is the [gcloud-ruby work in progress](http://quartzmo.github.io/gcloud-ruby/#/) in case you want to compare.

dwsupplee:
Awesome, thanks @quartzmo!

dwsupplee:
This is ready to merge as long as everything looks good.

stephenplusplus:
@jgeewax do we want to wait for any particular gcloud-php milestone before linking to it from the Node and Python docs?

jgeewax:
Yep - the main milestone is to get it so that composer command actually
works :)

(Even if it's 0.0.0.0.0.0.0.1... it must exist.)

dwsupplee:
@callmehiphop @stephenplusplus:

We've been asked to remove references to the term `gcloud` in the docs site for PHP. It looks to me the only site specific `gcloud` branding is the text on the top left in the header. Assuming that's correct, what do you guys think about conditionally modifying the header to look something like this?

screen shot 2016-03-27 at 11 09 33 am

stephenplusplus:
That's cool, but maybe we should all be removing the gcloud references? Any insight on this @jgeewax?

jgeewax:
If the library is called `gcloud`, we should keep those, but if not (ie, C#, Java, and PHP all use a hierarchical naming style), we should call it what it actually is.

And where we refer to the "project" or "library" in a non-technical sense, let's just call it what it is: the client library for Google Cloud Platform.

callmehiphop:
I don't see why the <kbd>manifest.json</kbd> file can't just list the title/graphic/etc. to show. Should be easy to implement.

dwsupplee:
The site defaults to the 'gcloud' custom type when initially loaded - which appears to be hard coded. It took me a bit of time to realize this is why the site wasn't loading for me, since I no longer have a `gcloud` type in the project.

@callmehiphop 
I'm thinking there may need to be a few additions to the manifest,

- Which custom type to load initially
- The name of the project (for situations like https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/service/service.html#L11)
- A header graphic as you mentioned

What do you think? I'm happy to help out and open a PR if that is helpful. If so, would you prefer I open a new one or add the work here?

callmehiphop:
@dwsupplee I think this PR should be fine :)

dwsupplee:
@callmehiphop 

I've updated the PR with a few new bells and whistles.

Added to manifest.json:
`libraryTitle` - The title for the library. This is used in the header once you land on the docs and in the overview toggle 'Getting started with...'.
`defaultService` - The service to load by default upon initial loading of the docs. 

Both additions are optional and will default to `gcloud` if not provided.

With this update swapping out `gcloud` with `Google Cloud Client Library` the overview toggle text sounds awkward reading as: `Getting Started with Google Cloud Client Library`.

Would changing the content to the following be fair?

`Google Cloud Client Library: Getting Started`

callmehiphop:
I think that's fine, or if we wanted to make it more future proof we could just rename it to simply `Getting Started`

README.md
```diff
@@ -5,6 +5,7 @@ Get up and running with your Google Cloud Platform client library.
   - Go - [gcloud-golang](https://github.com/googlecloudplatform/gcloud-golang)
   - Java - [gcloud-java](https://github.com/googlecloudplatform/gcloud-java)
   - Node.js - [gcloud-node](https://github.com/googlecloudplatform/gcloud-node)
+  - PHP - [gcloud-python](https://github.com/googlecloudplatform/gcloud-php)
```
callmehiphop: Link text here still says python

dwsupplee: Taken care of, thanks for the catch.

dwsupplee:
The content would be dynamic based on the `libraryTitle`, so it should be future proof either way. 

With that said, I prefer `Getting Started` :). Updates made, let me know what you think.

callmehiphop:
LGTM, are we still waiting for an official PHP release before merging this?

dwsupplee:
I think we'll want to make sure https://github.com/GoogleCloudPlatform/gcloud-php/pull/42 gets merged first, I'll let you know when that happens

dwsupplee:
@callmehiphop 

We just tagged release 0.3.0 and are ready for a merge whenever! :)

callmehiphop:
@dwsupplee Merged!

stephenplusplus:
WOOOO!!!

daspecster:
Nice!

jdpedrie:
awesome!

Issue #80, opened 2016-02-08
Add Method Groups


blowmage:
In gcloud-ruby we have a couple instances where we define a method on a class, and another method with the same name on instances of the class. Right now we cannot differentiate between the methods in the angular app. All methods, whether instance methods or static/class methods are listed in one group. We suggest nesting methods in under a group with a name, so that we can differentiate.

This would also help differentiate constructor methods. Instead of using the `constructor` attribute, we could create a "Constructor" group and list the constructor method under it.

Update: This issue will cover both method types (language-defined) and groups/categories (user-defined):
- [ ] Add `method.metadata.type` to support language-based method types, such as constructor, instance, and class/static
- [ ] Add `method.metadata.group` or `method.metadata.category` to support arbitrary, user-generated groups of methods, such as the "Data", "Lifecycle", and "Table" categories

callmehiphop:
SGTM, my only real question about this is how will it be laid out? Currently we provide a method list at the top of each API page, for users who want information about a particular method this feature is invaluable. How do we still provide this functionality while indicating the differences between various methods in a user friendly way?

blowmage:
What we've done is show the method lists by group. For example, see our current docs for [BigQuery's Dataset](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Bigquery/Dataset.html). We have multiple groups, of methods listed at the top of the page. And the page shows each group, and the methods contained in it, in order.

callmehiphop:
Ah, gotcha, something like that could definitely work. Were you planning on creating a method group for each of the categories listed in that table? Or would additional configurations be needed in order to map them properly.

blowmage:
One group per category.

callmehiphop:
Cool, historically we've always displayed this list in the constructor section, do you feel like that's still an appropriate place for it? Or that it would be better suited before the constructor but after the API title/overview?

blowmage:
This might be one of the ways in which ruby and javascript differ. We can place documentation on the class as well as the class's constructor method, so in ruby it makes more sense to list all methods after the class description. But in javascript the code construct is the constructor, so that gets more complicated.

blowmage:
Also, gcloud-ruby has many classes where the constructor is we intentionally marked as private so it doesn't get documented. We guide users to use other methods to create instances of those classes, and to not use the class constructor.

callmehiphop:
[...]

PR #100, opened 2016-02-23
Add method type to service template

quartzmo:
This PR addresses the potential for name conflicts between instance and class methods in Ruby. We actually have a number of methods in the `Gcloud` module that share the same name, so this problem is not hypothetical. See #80 for discussion.

Previously, page anchors were broken for these methods. In the demo for this PR, navigation to the `bigquery` [instance method](http://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud?method=bigquery) and  [class method](http://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud?method=bigquery-type-class) now both work.

I'm not really thrilled about the direction I took with this solution. I was hoping to introduce `type` as a parameter to the Services controller, but it would require more changes (to `DeeplinkService`, `customType`, etc)  than I was comfortable making without feedback/input. 

@callmehiphop I would be more than happy to throw away this PR if you can suggest or submit a better solution.

/cc @blowmage @stephenplusplus

### DIFF
--- a/site/src/app/docs/docs.service.js
+++ b/site/src/app/docs/docs.service.js
@@ -14,6 +14,8 @@
       if (method.metadata) {
         method.metadata.description = $sce.trustAsHtml(method.metadata.description);
         method.metadata.isConstructor = method.metadata.constructor === true;
+        method.metadata.anchor = getMethodAnchor(method.metadata);
+        method.metadata.typeSymbol = getMethodTypeSymbol(method.metadata.type);
 
         if (method.metadata.examples) {
           method.metadata.examples = method.metadata.examples.map(trustExample);
@@ -65,6 +67,22 @@
       param.description = $sce.trustAsHtml(param.description);
 
       return param;
+    }
+
+    function getMethodAnchor(metadata) {
+      var anchor = metadata.name;
+      if (metadata.type) {
+        anchor += "-type-" + metadata.type;
+      }
+      return anchor;
+    }
+
+    function getMethodTypeSymbol(type) {
+      var typeSymbol = "#";
+      if (type === "class") {
+        typeSymbol = ".";
+      }
+      return typeSymbol;
     }
 
     return {
--- a/site/src/app/service/service.html
+++ b/site/src/app/service/service.html
@@ -30,9 +30,9 @@
   </article>
   <article ng-repeat="method in service.methods">
     <h2 ng-if="method.metadata.isConstructor">{{::method.metadata.name}}</h2>
-    <h3 id="{{::method.metadata.name}}" ng-if="!method.metadata.isConstructor" class="method-heading">
-      <a class="permalink" ui-sref="docs.service({ method: method.metadata.name })">
-        <span>#</span>
+    <h3 id="{{::method.metadata.anchor}}" ng-if="!method.metadata.isConstructor" class="method-heading">
+      <a class="permalink" ui-sref="docs.service({ method: method.metadata.anchor })">
+        <span>{{::method.metadata.typeSymbol}}</span>
         {{::method.metadata.name}}
       </a>
     </h3>


callmehiphop:
I was afraid this day would come! Originally my fear was that we would add a language that supported overloading. While that's not what this PR aims to address, I think we could probably find a common solution for both.

Instead of adding verbiage to the ID to make them unique, could we just add a suffix in the form of a counter? e.g. `myMethodName-1`? That was my original plan for overloading.

quartzmo:
@callmehiphop The counter solution seems extremely flexible, and I can't imagine why it wouldn't work. I'll try it.

quartzmo:
@callmehiphop I couldn't imagine why it wouldn't work--until I tried it.

I blocked by how to create links to the correct method during the JSON build. This assumes that the suffix is generated by the site. Or did you think it would be in the JSON?

callmehiphop:
@quartzmo I thought the site would generate the links/ids. My thought process was that we would map the links in the service controller and as we found duplicates, we would add the appropriate suffix.

quartzmo:
@callmehiphop That is what I was thinking as well. So, how so I link to the class method (`Gcloud.bigquery`) instead of the instance method (`Gcloud#bigquery`) using the `data-custom-type` directive in our docs JSON?

blowmage:
This is why my suggestion was to assign every method having a unique ID, and as put that unique ID in the `data-custom-type` attribute.

callmehiphop:
@blowmage I was literally just going to ask if the IDs weren't unique!

quartzmo:
@callmehiphop By ID, do you mean the `name` field? Should I try to treat that as a unique ID? It should work as long as it isn't used for display. (Assuming the `title` field is always used for display. Why is `title` an array btw?)

blowmage:
In our case, we would assign the two `big query` methods different IDs. The ruby convention for class/static methods is a `.`, and an instance method is a `#`. So we would choose to use `Gcloud.bigquery` and Gcloud#bigquery`. Seeing those IDs the URL would feel natural for ruby users. it would also be easier for them to "hack" the URL and look or documentation without linking to it.

quartzmo:
However, currently, `name` is [used for display](https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/service/service.html#L32-L33) (I thought so):

blowmage:
In my opinion, I think the angular app's routing should be much simplified. It should be this:

/docs#/{version}/{identifier}`

And the `identifier` can be a guide, an object/class/service, or a method.

callmehiphop:
Would the `identifier` be present in the JSON?

blowmage:
Yeah, each entity would have the ID in the JSON.

quartzmo:
Should I abandon this PR then, in favor of simplified, ID-based routing? @callmehiphop will you be able to implement it fairly quickly, to resolve our issue of method name collision?

blowmage:
I think having an optional `type` value on a method entity in the JSON is a good thing. Similar to a `type` value on class/object/service entity that we can display to users so they use to differentiate between modules and classes in ruby. I would like this to be a thing that users are shown.

quartzmo:
@blowmage The code changes in this PR are mostly about navigation. Displaying metadata about types would be better presented in a new PR.

blowmage:
I have no opinion on using this PR or a different one. its up to you guys IMO. I was only trying to clarify my thoughts on why the routing becomes complicated without a unique ID on each entity in the JSON.

quartzmo:
I agree with @blowmage that displaying code object "type" to the user will improve the experience for Ruby programmers.

callmehiphop:
@blowmage so if the `identifier` is doubling up as an ID and URL (is that correct?) how would you display method? Unfortunately since this is an Angular app, the `#` is spoken for so I've been using query strings. Would it stay the same and it would be something like `/storage/bucket?method=delete`?

blowmage:
@callmehiphop In ruby we use `.` for class methods and `#` for instance methods. So the URLs I'd like to see is:

* `/gcloud-ruby/#/docs/v0.6.1/AUTHENTICATION` -> displays a guide
* `/gcloud-ruby/#/docs/v0.6.1/Gcloud/Bigquery` -> displays a service/class/code entity
* `/gcloud-ruby/#/docs/v0.6.1/Gcloud/Bigquery/Dataset#query` -> displays a method (instance method)
* `/gcloud-ruby/#/docs/v0.6.1/Gcloud.bigquery` -> displays an method (class method)

We would use ruby's standard notation of `Gcloud::Bigquery::Dataset#query`, but replace `::` with `/`.

I assume you can use `#` in the location hash of an angular app. I'm not aware of why angular would care if there were `#` characters in the location hash, as it should pull the location hash after the first `#` in the URL path. But I admit I am super ignorant of angular, so if that is a problem we would use a different character than `#` for the instance method IDs.

Another thing I like about this approach is gcloud-node can use IDs that will maintain the existing URLs. Set the method's ID to `storage/bucket?method=delete` and you are good to go.

callmehiphop:
@blowmage I think I actually tried using multiple hashes (with Angular) and ran into some routing issues previously. 

I think it would be ideal if all the languages followed a similar pattern for deeplinking. I'm not necessarily tied to using a query string, although I don't see many options here.

blowmage:
I'm sure we can work with any necessary character restrictions on a unique ID. If we are unable to use `#` then perhaps we could use `@`, or some other character that we are sure will lead to a unique ID.

`/gcloud-ruby/#/docs/v0.6.1/Gcloud/Bigquery/Dataset@query`

`...data retrieved from the <a custom-data-type="Gcloud/Bigquery/Dataset@query">Dataset#query</a> method...`

My point isn't to decide on what the unique IDs will be, only that they will be used.

quartzmo:
I agree with @blowmage, I think it will be fine for each platform to choose its own ID conventions (as long as they work with angular, of course.)

callmehiphop:
I agree that each platform choosing its own ID is ideal, my only concern was in regards to deeplinking based off of the IDs, but if you guys think you can make it work, that's fine with me.

quartzmo:
@callmehiphop After #102 is merged, I will update this PR for the removal of `metadata` and the change from `isConstructor` to `type`.

However, despite the commentary above (points noted), I still haven't come up with a better approach than the one I originally took, in which `type` is appended to the element id and anchor. The goal is to support linking to two methods with the same name. 

```js
anchor += "-type-" + metadata.type;
```

This approach meets the immediate needs of Ruby, but may not be ideal for all future use cases, such as overloaded methods (unless you want to set `type` to `overload-1`, `overload-2`,...).

So, any suggestions? Afaik, this is the final issue preventing gcloud-ruby from publishing our docs using the new site.

callmehiphop:
Well, if we have a proper `id` for each method now, could we not URL encode the id and set that as the value of `method`? e.g. `/docs/master/storage/bucket?method=storage%2Fbucket%23create`

quartzmo:
@callmehiphop Good thinking, let's do that!

quartzmo:
@callmehiphop For this to work, [usages of `method.name` for linking](https://github.com/callmehiphop/gcloud-common/blob/json-updates/site/src/app/service/service.html#L33-L34) need to be changed to url-encoded `method.id`.

Also, for our issue #80, we need to display the `method.type` in this section as well.

Where should we do this? PR #102?

callmehiphop:
@quartzmo I would say a new PR

quartzmo:
@callmehiphop ok will do!

quartzmo:
Replacing this outdated attempt with #107 and another upcoming PR for linking methods by `id`.

PR #107, opened 2016-03-18
Custom method type symbols

quartzmo:
This PR visibly differentiates methods by `type` with a custom, runtime-configured string (symbol or other label.) 

[refs #80]

The PR depends on a new (optional) `methodTypeSymbols` field in `manifest.json`, which maps display characters to method types. (For example: `#` for `instance`.)

This PR replaces outdated #100.

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -13,6 +13,7 @@
   * [`content`][content-key]
   * [`home`][home-key]
   * [`package`][package-key]
+  * [`methodTypeSymbols`][methodtypesymbols-key]
 * [Table of Contents][toc]
   * [`overview`][overview-key]
   * [`guides`][guides-key]
@@ -178,6 +179,25 @@ This allows you to specify what package manager you use and a direct link to you
     "title": "npm",
     "href": "https://www.npmjs.com/package/gcloud"
   }
+}
+```
+
+##### `methodTypeSymbols` key
+
+This optional list allows you to map language-specific method type names (such as `instance` or `static`) to labels or symbols displayed in the method listing (such as `#`.) If not provided, the pound sign (`#`) will be used for all method types.
+
+```js
+{
+  "methodTypeSymbols": [
+    {
+      "type": "class",
+      "symbol": "::"
+    },
+    {
+      "type": "instance",
+      "symbol": "#"
+    }
+  ]
 }
 ```
 
--- a/site/schemas/manifest.schema.json
+++ b/site/schemas/manifest.schema.json
@@ -57,6 +57,32 @@
         "title",
         "href"
       ]
+    },
+    "methodTypeSymbols": {
+      "id": "methodTypeSymbols",
+      "type": "array",
+      "description": "List of mappings from method types to display symbols",
+      "items": {
+        "id": "methodTypeSymbol",
+        "type": "object",
+        "description": "Maps a method type ('instance') to a symbol ('#')",
+        "properties": {
+          "type": {
+            "id": "type",
+            "type": "string",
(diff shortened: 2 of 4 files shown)


blowmage:
:+1:

quartzmo:
I will remove `don't merge` after I update `manifest.json.schema`.

quartzmo:
@callmehiphop I put the mappings from method type to display label in `manifest.json` (now documented in README and JSON schema), but I am totally open to another location.

site/src/app/components/custom-type/custom-type.directive.js
```diff
@@ -17,7 +17,7 @@
       var stateOptions = { inherit: false };
 
       if (parts.length > 1) {
-        stateParams.method = parts[1];
+        stateParams.method = decodeURIComponent(parts[1]);
```
callmehiphop: I think this needs to be updated to include the entire ID, `parts[1]` refers to the actual name of the method; `storage/bucket#create` yields `create`

quartzmo: This code works now (try some of the method links in this 
[Demo](http://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud)), however the method `id` values that I am currently using are not globally unique. The goal of this PR is limited to establishing support for method types. It's not about globally unique method ids.

quartzmo: I thought with this was to preserve the backward-compatible behavior of the `#` separator in the `customType` directive.

callmehiphop: Ah ok, I suppose that makes sense. I tested this locally against gcloud-node JSON and it broke deeplinking for us. Since the heading IDs were changed to use [`method.encodedId`](https://github.com/GoogleCloudPlatform/gcloud-common/pull/107/files/3c7d1401c01847ee36568109d3bd18fd1578dae3#diff-acc47ac7aa3d468f5644998af6237f2dR16), which is the full id, the old `customType` logic no longer works.

quartzmo: Oh shoot. Did you test against gcloud-node JSON that uses IDs like this?

```js
id: "gcloud#bigquery",
name: "bigquery",
type: "instance",
```

If so, changing the `id` to match the `name` (`"bigquery"`) should work. In gcloud-ruby JSON for this PR, the `id` contains the symbol indicating the method type (`"#bigquery"`).

But I realize I am basically highjacking the `id` field for a purpose other than global identification of the code object. I think the best approach would require a more ambitious overhaul of `DeeplinkService` and `customType`.

quartzmo: I think the root problem is the logic in `customType` that splits the string with `#` to get the service name and the method name. I'll see if I can come up with another solution. Please let me know if you have any ideas!

callmehiphop: > But I realize I am basically highjacking the id field for a purpose other than global identification of the code object. I think the best approach would require a more ambitious overhaul of DeeplinkService and customType.

I agree!

quartzmo: OK I'll work on it.

quartzmo:
@callmehiphop I have removed all navigation and linking changes from this PR in order to keep it focused on just the display of custom method type symbols. PTAL and if good, please merge. I will continue work on a solution for linking to full-path method `id`s such as `storage/bucket#create` in a new PR.

Issue #79, opened 2016-02-08
Add Unique ID to Guides, Code Constructs, and Methods


blowmage:
The current angular app displays three main entities, a guide, code construct, and a code method. The routing rules to link to any of these entities with the `data-custom-type` attribute can get complicated quickly. We would prefer that each entity has a unique ID, and that the URL routing and `data-custom-type` attributes use this ID.

quartzmo:
One motivation behind this idea is to facilitate linking to deeply nested code types that are not listed in the navigation. #78 is therefore likely to be a prerequisite.

blowmage:
Agreed. :+1:

callmehiphop:
:+1:

callmehiphop:
Closed by #109 and #115

PR #109, opened 2016-03-19
site: support links to method ids

quartzmo:
This PR allows deep linking to a method using the method's `id` rather than its name, which was problematic in Ruby due to duplicate method names in the same namespace.

[refs #79]

The PR depends on adding entries to `types.json` for all methods. The entries for methods should reference the `title` and `contents` values for the parent class (or other namespace). For example:

```js
  {
    "id": "gcloud",
    "title": "Gcloud",
    "contents": "gcloud.json"
  },
  {
    "id": "gcloud.new",
    "title": "Gcloud",
    "contents": "gcloud.json"
  },
  {
    "id": "gcloud#datastore",
    "title": "Gcloud",
    "contents": "gcloud.json"
  },
  ...
  {
    "id": "gcloud.datastore",
    "title": "Gcloud",
    "contents": "gcloud.json"
  },
  ...
```

This PR replaces outdated #100.

### DIFF
--- a/site/src/app/components/custom-type/custom-type.directive.js
+++ b/site/src/app/components/custom-type/custom-type.directive.js
@@ -6,27 +6,17 @@
     .directive('customType', customType);
 
   /** @ngInject */
-  function customType($state, manifest) {
+  function customType($state) {
     var convertToServicePath = function(customType) {
-      var parts = customType.split('#');
       var stateName = 'docs.service';
       var stateParams = {
-        serviceId: parts[0],
+        serviceId: customType,
         version: $state.params.version
       };
       var stateOptions = { inherit: false };
 
-      if (parts.length > 1) {
-        stateParams.method = parts[1];
-      }
-
       return $state.href(stateName, stateParams, stateOptions);
     };
-
-    var converter = {
-      node: convertToServicePath,
-      ruby: convertToServicePath
-    }[manifest.lang];
 
     return {
       restrict: 'A',
@@ -36,7 +26,7 @@
           elem.html(customType); // Set path as text if no text in element
         }
         elem.addClass('skip-external-link')
-          .attr('href', converter(customType.replace('[]', '')));
+          .attr('href', convertToServicePath(customType.replace('[]', '')));
       }
     };
   }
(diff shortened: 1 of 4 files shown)


quartzmo:
I've noticed in the demo that when navigating to methods, the correct item in the left nav does not get highlighted. I'll try to correct this bug as soon as I can.

quartzmo:
Momentarily placing this in `don't merge`. I'm rebasing on #107 and looking into the issue with left nav active highlight.

quartzmo:
@callmehiphop I've removed `don't merge` and believe this is ready to merge. PTAL!

You can review the behavior in the gcloud-ruby demo, in which [/docs/master/gcloud.bigquery](https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud.bigquery) and [/docs/master/gcloud#bigquery](https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud#bigquery) now both link correctly to the class and instance methods (respectively) of the same name.

callmehiphop:
@quartzmo looks good! My only real complaint is that clicking a link to a method on the same page reloads the json/view, but I fear preventing that would require a much larger and more complicated refactor.

quartzmo:
@callmehiphop Agreed, and sorry that this solution involves that additional inefficiency. If you don't mind merging this PR to get gcloud-ruby past its lone remaining blocker, I would be happy to open an issue for this unneeded load, and to keep working on improving the internals.

callmehiphop:
@quartzmo I don't mind at all. I was just looking through the site some more and I noticed that https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud#bigquery does not deeplink properly - however https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud%23bigquery does. Is that by design?

quartzmo:
@callmehiphop It was by design, based on what you commented [here](https://github.com/GoogleCloudPlatform/gcloud-common/pull/100#issuecomment-188850746):

> Unfortunately since this is an Angular app, the # is spoken for so I've been using query strings.

In addition, the `uiSref` and/or `a` directive appears to encode links (`#` to `%23`) [in the template](https://github.com/quartzmo/gcloud-common/blob/method-links/site/src/app/service/service.html#L34), so I just went with it.

callmehiphop:
@quartzmo got it, I'm going to merge then!

quartzmo:
@callmehiphop Thanks! Do you want to open an issue to avoid reloading the JSON/view when navigating to an "anchor" in the same page, or should I?

callmehiphop:
I can do that!

PR #115, opened 2016-03-24
site: revert method URLs to ?method={id}

### LINKED ISSUE #113: Switch "%23{name}" back to "?method={name}"
Is it possible to remove the "%23" and bring back "?method=" in our URLs?

quartzmo:
Like PR #109, which is partially reverts, this PR allows deep linking to a method using the method's `id` rather than its name, which was problematic in Ruby due to duplicate method names in the same namespace.

However, unlike #109, this PR preserves the `?method={id}` query param URL signature. Projects without method name conflicts, such as gcloud-node, may simply use the same value for `method.id` as for `method.name` in docs JSON. (gcloud-ruby creates `method.id` by appending `method.type` to `method.name`.)

[closes #113]

/cc @blowmage @stephenplusplus

### DIFF
--- a/site/src/app/components/custom-type/custom-type.directive.js
+++ b/site/src/app/components/custom-type/custom-type.directive.js
@@ -7,13 +7,17 @@
 
   /** @ngInject */
   function customType($state) {
-    var convertToServicePath = function(customType) {
+    var convertToServicePath = function(customType, method) {
       var stateName = 'docs.service';
       var stateParams = {
         serviceId: customType,
         version: $state.params.version
       };
       var stateOptions = { inherit: false };
+
+      if (method) {
+        stateParams.method = method;
+      }
 
       return $state.href(stateName, stateParams, stateOptions);
     };
@@ -22,11 +26,14 @@
       restrict: 'A',
       link: function (scope, elem, attrs) {
         var customType = attrs.customType;
+        var method = attrs.method;
+
         if (elem.html().length === 0) {
-          elem.html(customType); // Set path as text if no text in element
+          // Set narrowest scope as text if no text in element
+          elem.html(method ? method : customType);
         }
         elem.addClass('skip-external-link')
-          .attr('href', convertToServicePath(customType.replace('[]', '')));
+          .attr('href', convertToServicePath(customType.replace('[]', ''), method));
       }
     };
   }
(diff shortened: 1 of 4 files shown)


site/src/app/components/custom-type/custom-type.directive.js
```diff
@@ -22,11 +26,13 @@
       restrict: 'A',
       link: function (scope, elem, attrs) {
         var customType = attrs.customType;
+        var method = attrs.method;
+
         if (elem.html().length === 0) {
           elem.html(customType); // Set path as text if no text in element
```
callmehiphop: I think this needs to change to something like

```javascript
elem.html(customType + method);
```

Which means we also might want to set method to an empty string when falsey.

quartzmo: @callmehiphop Good catch! I didn't notice while testing manually because we don't have any such empty links (our JSON generation already does something similar that guarantees text).

It's a bit of a judgement call as to what the generated text should look like. Can you give me your preferred solution for this and I'll update the PR?

callmehiphop: hmm, how about we if the method is available we dump that in otherwise if it's missing we just dump the module in?

```js
elem.html(method ? method : customType);
```

quartzmo: SGTM! Updating!

callmehiphop:
Thanks @quartzmo for all your contributions!

stephenplusplus:
Thanks @quartzmo!

quartzmo:
@callmehiphop @stephenplusplus Thanks you guys! We are getting very close! I think the only remaining blocking issue is to transition our article/overview code examples to Highlight.js. I think this may require a change to apply Highlight.js to code blocks in descriptions? If so, I'll open a new issue for it.

stephenplusplus:
@callmehiphop can our site just apply hljs to any `<pre>` elements it finds? Maybe an Angular directive `<pre>` (to extend the real `<pre>`)?

callmehiphop:
> can our site just apply hljs to any <pre> elements it finds? Maybe an Angular directive <pre> (to extend the real <pre>)?

That would work, I put some similar logic in when parsing markdown - https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/site/src/app/components/markdown/markdown.directive.js#L17. I don't see why we couldn't extract it and overload the `<pre>` tag.

PR #117, opened 2016-03-25
docs: overload pre tag to apply syntax highlighting

### LINKED ISSUE #116: Create Angular directive: <pre> to auto-apply Highlight.js
Carrying on from https://github.com/GoogleCloudPlatform/gcloud-common/pull/115#issuecomment-200984631

With Angular, we should overload the `<pre>` element so that any code within it is automatically syntax highlighted. This should apply to snippets from any source (examples, descriptions, Markdown guides, etc.).

Will hljs know what language to use for highlighting from its own global configuration or should we add this functionality in manually through our own app config?

callmehiphop:
Closes #116 

This PR will search for code snippets via `<pre><code></code></pre>` tags and auto highlight them. It's possible to bypass this by adding a class of `skip-highlight` to the `<pre>` tag. If the `<pre>` does **not** contain a `<code>` tag, it will **not** apply highlighting.

/cc @quartzmo @stephenplusplus

### DIFF
--- a/site/src/app/components/markdown/markdown-highlight.directive.js
+++ b/site/src/app/components/markdown/markdown-highlight.directive.js
@@ -1,0 +1,33 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('gcloud')
+    .directive('pre', markdownHighlight);
+
+  /** @ngInject */
+  function markdownHighlight(hljs) {
+    return {
+      restrict: 'E',
+      link: function(scope, elem) {
+        if (elem.hasClass('skip-highlight')) {
+          return;
+        }
+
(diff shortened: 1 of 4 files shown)


stephenplusplus:
> If the <pre> does not contain a <code> tag, it will not apply highlighting.

Why not?

callmehiphop:
IIRC that's actually a mechanism of hljs to only highlight code inside of `<code>` tags.

stephenplusplus:
Hmm, I guess it is. That seems weird. From an HTML standpoint, I thought `code` was inline and `pre` was block. Oh well!

quartzmo:
@callmehiphop Awesome. Thank you!

PR #119, opened 2016-03-25
docs: limit scope of highlight languages

callmehiphop:
Looks like `hljs` is misinterpreting some of our snippets ([`startQuery` as CSS](https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.29.0/bigquery?method=startQuery)).

This will explicitly set all code examples on service pages to the code flavor defined in the <kbd>manifest.json</kbd>

### DIFF
--- a/site/src/app/service/service.html
+++ b/site/src/app/service/service.html
@@ -70,7 +70,7 @@
       <div ng-repeat="example in method.examples">
         <div ng-if="example.caption" bind-html-compile="example.caption"></div>
         <div ng-if="example.code" class="code-block">
-          <pre><code class="hljs" bind-html-compile="example.code"></code></pre>
+          <pre><code class="hljs {{::markdown}}" bind-html-compile="example.code"></code></pre>
         </div>
       </div>
     </section>


stephenplusplus:
Those are good ones to add, but is it possible to not make hljs guess at all and just default to the language of our library, unless we explicitly add a class in our markup?

callmehiphop:
Yep! Totally possible by adding the language as a class. However, since the class names themselves don't use a prefix of any kind, we would either need to whitelist the languages to check for or include additional custom attributes letting us know a language is present.

quartzmo:
@callmehiphop @stephenplusplus Great! Thank you both so much.

@blowmage @jgeewax I've redeployed the gcloud-ruby [Demo](http://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud) with the latest site and JSON. Everything we found this week should be fixed.

quartzmo:
@callmehiphop Actually, it looks like the Ruby highlighting is still inconsistent, although both examples appear to be styled by Highlight.js

ruby_style

Inspecting the page, I see the upper example to have `<code class="hljs lua" ...`, the lower to have  `<code class="hljs stylus" ...`.

(There is a chance that gh-pages is still serving the site version previous to this PR and just needs to propagate my latest push, but I thought I'd better share this.)

callmehiphop:
@quartzmo I don't think your demo has the latest changes yet. I did a quick search for the updated code and I did not find it.

quartzmo:
@callmehiphop OK, great. I'll post again here once I've sorted it out.

quartzmo:
@callmehiphop The demo is updated and looking good! Thank you again for your fast work on this today! @blowmage and I really appreciate your help so much.

callmehiphop:
@quartzmo no worries! Glad it's looking good :+1:

PR #134, opened 2016-05-05
site: add support for configurable title characters

### LINKED ISSUE #133: Docs site generated navigation titles
In the docs site there's a convention of `Thing » Subthing`. This looks great, but it might better maintain consistency if it was managed on the site generation side rather than allowing projects to change the convention?

Mentioned: https://github.com/GoogleCloudPlatform/gcloud-python/pull/1775#discussion_r62203856

callmehiphop:
Closes #133 

This adds an (optional) field to the <kbd>manifest.json</kbd> called `titleDelimiter`. If this is set and a page title is found in the form of an Array, the Angular app will join said array with the value of `titleDelimiter`.

---

### Example

<kbd>manifest.json</kbd>

```js
{
  "titleDelimiter": "::"
}
```

<kbd>types.json</kbd>

```js
[{
  "id": "storage/bucket",
  "title": ["Storage", "Bucket"],
  "contents": "storage/bucket.json"
}]
```

Yields `Storage::Bucket` for the title of `/docs/{version}/storage/bucket`.

/cc @jgeewax @daspecster

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -9,6 +9,7 @@
   * [`lang`][lang-key]
   * [`friendlyLang`][friendlylang-key]
   * [`markdown`][markdown-key]
+  * [`titleDelimiter`][delimiter-key]
   * [`versions`][versions-key]
   * [`content`][content-key]
   * [`home`][home-key]
@@ -126,6 +127,16 @@ The Angular app currently leverages a code highlighting library called [highligh
 ```js
 {
   "markdown": "javascript"
+}
+```
+
+##### `titleDelimiter` key
+
+*This key is completely optional*. Used for joining title strings together, if omitted the default value is `" » "`.
+
+```js
+{
+  "titleDelimiter": " » "
 }
 ```
 
@@ -274,7 +285,7 @@ The types section lists all the available data types for the application. All ty
 
 The `id` key will be used to map content from urls - `/docs/v0.28.0/storage/bucket` will map to `"id": "storage/bucket"`
 
-The `title` key will be used to create a title on the service page.
+The `title` key will be used to create a title on the service page. You can provide either a string or an array of strings that will be joined by a special character specified in the <kbd>manifest.json</kbd> via [`titleDelimiter`][delimiter-key].
 
 ```js
 {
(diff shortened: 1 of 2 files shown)


site/README.md
```diff
@@ -129,6 +130,16 @@ The Angular app currently leverages a code highlighting library called [highligh
 }
 ```
 
+##### `titleDelimiter` key
+
+*This key is completely optional*. Used for joining title strings together, if omitted the default value is `" » "`.
```
daspecster: Not sure the best way to phrase this, but is there is a way to indicate in the docs that this is actually the language specific namespace delimiter?

site/README.md
```diff
@@ -274,7 +285,7 @@ The types section lists all the available data types for the application. All ty
 
 The `id` key will be used to map content from urls - `/docs/v0.28.0/storage/bucket` will map to `"id": "storage/bucket"`
 
-The `title` key will be used to create a title on the service page.
+The `title` key will be used to create a title on the service page. You can provide either a string or an array of strings that will be joined by a special character specified in the <kbd>manifest.json</kbd> via [`titleDelimiter`][delimiter-key].
```
daspecster: Maybe something like "Titles are generally representations of the language namespace."

site/README.md
callmehiphop: I can add some extra verbiage stating that, however I think saying it's for namespacing is misleading because in the case of Node, it doesn't represent our API hierarchy. We're actually just blanketing a concept. e.g. `Storage » File` represents how to deal with Files in the Storage API, however the actual hierarchy is `Storage » Bucket » File`

site/README.md
callmehiphop: RE: https://github.com/GoogleCloudPlatform/gcloud-common/pull/134#discussion_r62230405

site/README.md
daspecster: I'm good with this. I guess I'm just trying to think of a way to make it really clear since it's in `types.json`.  I don't have any good suggestions for it right now though.

daspecster:
👍

PR #146, opened 2016-06-13
add margin between examples with no captions

dwsupplee:
If an example has no caption it merges into the previous example making it look as if there is one big example.

This PR adds a margin between examples for these instances.

I added a new class `margin-top` which follows the standard for other utility classes. What do guys you think, does this seem useful?

screen shot 2016-06-13 at 12 08 16 pm

### DIFF
--- a/site/src/app/service/service.html
+++ b/site/src/app/service/service.html
@@ -31,7 +31,7 @@
       <h4>Example</h4>
       <div ng-repeat="example in service.examples">
         <div ng-if="example.caption" bind-html-compile="example.caption"></div>
-        <div ng-if="example.code" class="code-block">
+        <div ng-if="example.code" class="code-block" ng-class="{'margin-top': !example.caption && $index > 0}>
           <pre><code class="hljs {{::markdown}}" bind-html-compile="example.code"></code></pre>
         </div>
       </div>
@@ -77,7 +77,7 @@
       <h4>Example</h4>
       <div ng-repeat="example in method.examples">
         <div ng-if="example.caption" bind-html-compile="example.caption"></div>
-        <div ng-if="example.code" class="code-block">
+        <div ng-if="example.code" class="code-block" ng-class="{'margin-top': !example.caption && $index > 0}">
           <pre><code class="hljs {{::markdown}}" bind-html-compile="example.code"></code></pre>
         </div>
       </div>
--- a/site/src/app/styles/_main.scss
+++ b/site/src/app/styles/_main.scss
@@ -255,6 +255,10 @@ ul {
     margin: 10px;
 }
 
+.margin-top {
+   margin-top: 10px;
+}
+
 .margin-left {
     margin-left: 10px;
 }


callmehiphop:
I might be alone in this, but in the case of multiple examples I think I would expect a caption to explain to me what sets this particular snippet apart from the other examples (otherwise I would assume 1 giant example).

What's the use case for multiple examples with no captions?

jgeewax:
I'm with @callmehiphop : We should have *some* caption describing what the example is, right?

dwsupplee:
The screenshot above is a use case. One example is a simple call, the second is passing in more complex options. Sure we could add `And now... WITH MORE OPTIONS!` or whatnot as a caption, but it seems odd to not provide any spacing since captions are optional.

jgeewax:
I'd expect something like...

---

Upload an image:

```php
$bucket->upload...
```

Upload an image with extra options:

```php
$bucket->upload ( $options...)
```

callmehiphop:
I think in node's case, we usually just mark the parameter as optional and document it verbosely so it would look more like

```php
$options = [
  'resumable' => true
];

$bucket->upload(
  $options, 
  fopen('image.jpg', 'r')
);
```

stephenplusplus:
In this case, I would try to frame the examples around why a user would want to use those options. So, turn the example that shows "here's how the object we expect looks" (param docs should lay that out) into "here's how to set the contentType", since that or something like it is probably why the user landed on the docs page.

dwsupplee:
I think there could be any number of use cases.

My point is since the caption is optional we should assume there will be cases we may accidentally forget to add it or feel like it adds verbosity - and why wouldn't we cover ourselves to make the docs more clear in those situations?

callmehiphop:
Captions should really only be considered optional for the first example in a list of examples. I think the user is always going to assume the first example will cover the most common/simple use case. I would expect every example afterward to contain a caption explaining what sets this example apart from the common use case.

stephenplusplus:
Maybe we should we make all captions after the first example required?

quartzmo:
> Maybe we should we make all captions after the first example required?

This has been our style in gcloud-ruby, so sounds good to me.

dwsupplee:
I like that idea @stephenplusplus, I'm just not sure there is a way to validate the first item can be optional while the rest are required with json schema.

daspecster:
@dwsupplee this looks like it might be useful but doesn't look available yet.
JSON Schema Switch Proposal: https://github.com/json-schema/json-schema/wiki/switch-(v5-proposal)

stephenplusplus:
@dwsupplee are you using a json-schema linter/validator in your build process?

dwsupplee:
@stephenplusplus Not yet, but it's high on the list of things to get done.

@daspecster thanks for the links! It seems like those are pretty close to what we need, but unless I'm misunderstanding something it doesn't validate the order.

After a little digging it looks like we could use the `additionalItems` property.

```js
"examples": {
  "id": "examples",
  "type": "array",
  "description": "List of examples showing how the method is used",
  "items": [
    {
      "id": "example",
      "type": "object",
      "description": "Object describing an example",
      "properties": {
        "caption": {
          "id": "caption",
          "type": "string",
          "description": "Caption explaining what the example is attempting to accomplish"
        },
        "code": {
          "id": "code",
          "type": "string",
          "description": "Code snippet showing how the method can be used"
        }
      }
    }
  ],
  "additionalItems": {
    "id": "example",
    "type": "object",
    "description": "Object describing an example",
    "properties": {
      "caption": {
        "id": "caption",
        "type": "string",
        "description": "Caption explaining what the example is attempting to accomplish"
      },
      "code": {
        "id": "code",
        "type": "string",
        "description": "Code snippet showing how the method can be used"
      }
    },
    "required": ["caption", "code"]
  }
}
```

Would we prefer I close this out and instead open a PR modifying the schema?

callmehiphop:
@dwsupplee I think that's totally the way to go!

PR #157, opened 2016-07-29
site: modularization support

### LINKED ISSUE #52: docs: individual module version management
It seems like there's been a bit of chatter about some libraries moving toward individual modules instead of a single giant module. @blowmage [made a good point](https://github.com/GoogleCloudPlatform/gcloud-common/issues/49#issuecomment-173008535) about how versioning the individual modules will complicate the docs app.

The simplest way to solve that problem (that I can think of) might be to include a menu that simply allows one to switch between the main gcloud module and all individual modules. I put together a quick comp to illustrate the idea (forgive me if it's hideous).

default-state
expanded-state

Switching modules could repopulate the version dropdown to only contain the available versions for the selected module. The API section would probably only be populated with relating pages, for example, when selecting gcloud-bigquery, only Dataset, Job and Table would be listed.

Thoughts?

callmehiphop:
Closes #52 

This PR aims to add support for multiple modules while also upholding the old way of doing docs.

### manifest.json

The big change here is the addition of the `modules` key, this will be an array of modules that can be viewed. Each module should include a default service to be routed to as well as all available versions of said module.

You can specify the default module with the new `defaultModule` key.

```json
{
  "lang": "node",
  "friendlyLang": "Node.js",
  "markdown": "javascript",
  "titleDelimiter": "::",
  "defaultModule": "google-cloud",
  "moduleName": "gcloud-node",
  "content": "json",
  "home": "home.html",
  "package": {
    "title": "npm",
    "href": "https://www.npmjs.com/package/gcloud"
  },
  "modules": [{
    "id": "google-cloud",
    "name": "google-cloud",
    "defaultService": "gcloud",
    "versions": [
      "v0.29.0",
      "v0.10.0",
      "master"
    ]
  }, {
    "id": "bigquery",
    "name": "google-cloud-bigquery",
    "defaultService": "bigquery",
    "versions": [
      "v0.10.0",
      "master"
    ]
  }]
}
```

You'll also notice a `moduleName` key - this is for when `gcloud` goes away. Rather than the app trying to guess your repo name/stackoverflow tag, I thought it would be best to provide it in the manifest.json.

### Folder Hierarchy 

To support multiple modules a folder hierarchy change was needed, here is a rough example of how the JSON files should be laid out.

```sh
 ├─── google-cloud
 |     └─── master
 |           ├─── index.json
 |           ├─── types.json
 |           ├─── toc.json
 |           └─── vision
 |                 └─── index.json
 └─── vision
       ├─── master
       |     ├─── index.json
       |     ├─── types.json
       |     └─── toc.json
       └─── v0.2.0
             ├─── index.json
             ├─── types.json
             └─── toc.json
```

The obvious thing here being that since each version of each module should viewable, that means we'll need a table of contents and a list of available types per version.

### URL Changes

Previously all the links in our docs resembled something like

`/gcloud-{lang}/#/docs/{version}/{module}/{object}`

Now it will look like

`/gcloud-{lang}/#/docs/{module}/{version}/{object}`

## TODO

- [x] Update JSON Schema
- [x] Add routing support for individual modules
- [x] Documentation
- [x] Linking to modules via `custom-type` directive
- [x] Dropdown directive
- [x] Remove `gcloud` references

## Demo

Source files - https://github.com/callmehiphop/gcloud-common-modular
[Demo](https://callmehiphop.github.io/gcloud-common-modular/#/docs/google-cloud/v0.29.0/gcloud)

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -17,6 +17,9 @@
   * [`home`][home-key]
   * [`package`][package-key]
   * [`methodTypeSymbols`][methodtypesymbols-key]
+  * [`moduleName`][modulename-key]
+  * [`defaultModule`][defaultmodule-key]
+  * [`modules`][modules-key]
 * [Table of Contents][toc]
   * [`overview`][overview-key]
   * [`guides`][guides-key]
@@ -37,6 +40,7 @@
   * [Landing Page][landing-page]
   * [Overview Section][overview-section]
   * [Custom Page Headers][custom-page-headers]
+* [Modular Docs][modular-docs]
 * [Deploying][deploying]
 
 ## Installation
@@ -229,6 +233,49 @@ This optional list allows you to map language-specific method type names (such a
     {
       "type": "instance",
       "symbol": "#"
+    }
+  ]
+}
+```
+
+##### `moduleName`
+
+The name of your module, this is used to create links to GitHub/Stackoverflow/etc.
+
+```js
+{
+  "moduleName": "gcloud-node"
+}
+```
+
+##### `defaultModule`
+
+When using [modular docs][modular-docs], use this key to set the default module. The app will direct the user to this module in the event of an error and/or when the module has been omitted altogether.
+
+The value should be the `id` of a module listed within the [`modules`][modules-key] array.
+
+```js
+{
+  "defaultModule": "google-cloud"
+}
+```
+
+##### `modules`
+
+This field is only needed when using modular docs. When present it should contain a list of separate modules you wish to display docs for.
+
+```js
+{
+  "modules": [
+    {
+      "id": "google-cloud", // id used for routing
+      "name": "google-cloud", // name used for display purposes
+      "defaultService": "storage", // default service page
+      "versions": [ // available versions of the module
+        "v0.29.0",
+        "v0.10.0",
+        "master"
+      ]
     }
   ]
 }
@@ -510,6 +557,29 @@ See [gcloud-node docs][gcloud-node-docs] for a working demo. The contents of the
 
 See a living example of gcloud-node's overview template [here][gcloud-node-overview].
 
+## Modular Docs
+
+If your library uses a modular approach (a separate installable package per service), you can optionally leverage the modular docs. To do this you must first make sure that your [manifest][manifest] contains the [`modules`][modules-key] and [`defaultModule`][defaultmodule-key] fields.
+
+After which, you must create a folder within your content directory for each package. Each package should map to a module listed within your manifest. Here is an example of what your content folder should resemble
+
+```sh
+json
+ ├─── bigquery #standalone package
+ |     └─── master
+ |           ├─── toc.json # table of contents for this version
+ |           ├─── types.json # list of custom types for this version
+ |           └─── index.json # # one file per class/module/etc.
+ |
+ └─── google-cloud # bundled package
+       └─── v0.28.0
+             ├─── toc.json # table of contents for this version
+             ├─── types.json # list of custom types for this version
+             └─── storage # one folder per service
+                   ├─── index.json 
+                   └─── bucket.json
+```
+
 ## Deploying
 
 Integrating the shared docs app is fairly painless. There is a [shell script][gcloud-common-deploy] that will push the updated site code to your `gh-pages` after a successful merge. To enable this, simply append a command to push to your repo to the script.
@@ -539,6 +609,9 @@ Please refer to gcloud-node's [`gh-pages` branch][gcloud-node-ghpages] for an ex
 [guides-key]: #guides-key
 [services-key]: #services-key
 [package-key]: #package-key
+[modulename-key]: #modulename-key
+[defaultmodule-key]: #defaultmodule-key
+[modules-key]: #modules-key
 [json-schema]: #json-docs-schema
 [id-key]: #id-key
 [description-keys]: #description-and-caption-keys
@@ -558,6 +631,7 @@ Please refer to gcloud-node's [`gh-pages` branch][gcloud-node-ghpages] for an ex
 [types]: #types
 [types-key]: #types-key
 [methodtypesymbols-key]: #methodtypesymbols-key
+[modular-docs]: #modular-docs
 
 [nodejs]: https://nodejs.org/en/
 [hljs]: https://highlightjs.org/
--- a/site/schemas/manifest.schema.json
+++ b/site/schemas/manifest.schema.json
@@ -22,6 +22,16 @@
       "type": "string",
       "description": "Custom service type to load by default."
     },
+    "moduleName": {
+      "id": "moduleName",
+      "type": "string",
+      "description": "Name of the module used to link to Github repo"
+    },
+    "defaultModule": {
+      "id": "defaultModule",
+      "type": "string",
+      "description": "id of default module when using modular docs"
+    },
     "markdown": {
       "id": "markdown",
       "type": "string",
@@ -93,15 +103,68 @@
           "symbol"
         ]
       }
+    },
+    "modules": {
+      "id": "modules",
+      "type": "array",
+      "description": "List of modules and their versions",
+      "items": {
+        "id": "module",
+        "type": "object",
+        "description": "Information for individual module",
+        "properties": {
+          "id": {
+            "id": "id",
+            "type": "string",
+            "description": "id used within route parameter"
+          },
+          "name": {
+            "id": "name",
+            "type": "string",
+            "description": "Friendly name used for module switching dropdown"
+          },
+          "defaultService": {
+            "id": "defaultService",
+            "type": "string",
+            "description": "The id of the default service for the module"
+          },
+          "versions": {
+            "id": "versions",
+            "type": "array",
+            "description": "An array of all available versions for this module",
+            "items": {
+              "id": "version",
+              "type": "string",
+              "description": "The semver version string"
+            }
+          }
+        },
+        "required": [
+          "id",
+          "name",
+          "defaultService",
+          "versions"
+        ]
+      }
     }
   },
   "required": [
     "lang",
     "friendlyLang",
+    "moduleName",
     "markdown",
-    "versions",
     "content",
     "home",
     "package"
-  ]
+  ],
+  "oneOf": [{
+    "required": [
+      "versions"
+    ]
+  }, {
+    "required": [
+      "defaultModule",
+      "modules"
+    ]
+  }]
 }
--- a/site/schemas/service.schema.json
+++ b/site/schemas/service.schema.json
@@ -17,6 +17,11 @@
       "type": "string",
       "description": "Brief overview of the service"
     },
+    "overview": {
+      "id": "overview",
+      "type": "string",
+      "description": "Replacement for overview.html file"
+    },
     "resources": {
       "id": "resources",
       "type": "array",
--- a/site/src/app/components/custom-type/custom-type.directive.js
+++ b/site/src/app/components/custom-type/custom-type.directive.js
@@ -11,7 +11,8 @@
       var stateName = 'docs.service';
       var stateParams = {
         serviceId: customType,
-        version: $state.params.version
+        version: $state.params.version,
+        package: $state.params.package || ''
       };
       var stateOptions = { inherit: false };
 
--- a/site/src/app/components/custom-type/custom-type.directive.js
+++ b/site/src/app/components/custom-type/custom-type.directive.js
@@ -6,21 +6,39 @@
     .directive('customType', customType);
 
   /** @ngInject */
-  function customType($state) {
-    var convertToServicePath = function(customType, method) {
-      var stateName = 'docs.service';
-      var stateParams = {
-        serviceId: customType,
-        version: $state.params.version,
-        module: $state.params.module || ''
-      };
-      var stateOptions = { inherit: false };
+  function customType($state, util) {
+    var convertToServicePath = function(scope, customType, method) {
+      var isInclusive = !!util.findWhere(scope.docs.types, {
+        id: customType
+      });
 
-      if (method) {
-        stateParams.method = method;
+      var href = [
+        '#',
+        'docs'
+      ];
+
+      if (isInclusive) {
+        var params = [
+          $state.params.version,
+          customType
+        ];
+
+        if ($state.params.module) {
+          params.unshift($state.params.module);
+        }
+
+        href = href.concat(params);
+      } else {
+        href.push(customType);
       }
 
-      return $state.href(stateName, stateParams, stateOptions);
+      href = href.join('/');
+
+      if (method) {
+        href += '?method=' + method;
+      }
+
+      return href;
     };
 
     return {
@@ -33,8 +51,9 @@
           // Set narrowest scope as text if no text in element
           elem.html(method ? method : customType);
         }
-        elem.addClass('skip-external-link')
-          .attr('href', convertToServicePath(customType.replace('[]', ''), method));
+
+        var href = convertToServicePath(scope, customType.replace('[]', ''), method);
+        elem.addClass('skip-external-link').attr('href', href);
       }
     };
   }
--- a/site/src/app/components/custom-type/custom-type.directive.js
+++ b/site/src/app/components/custom-type/custom-type.directive.js
@@ -12,7 +12,7 @@
       var stateParams = {
         serviceId: customType,
         version: $state.params.version,
-        package: $state.params.package || ''
+        module: $state.params.module || ''
       };
       var stateOptions = { inherit: false };
 
--- a/site/src/app/components/dropdown/dropdown.directive.js
+++ b/site/src/app/components/dropdown/dropdown.directive.js
@@ -1,0 +1,30 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('gcloud')
+    .directive('dropdown', dropdown);
+
+  /** @ngInject */
+  function dropdown() {
+    return {
+      restrict: 'E',
+      replace: true,
+      templateUrl: 'app/components/dropdown/dropdown.html',
+      transclude: true,
+      scope: {
+        selected: '='
+      },
+      controller: DropdownCtrl,
+      controllerAs: 'dropdown',
+      bindToController: true
+    };
+  }
+
+  /** @ngInject */
+  function DropdownCtrl() {
+    var dropdown = this;
+
+    dropdown.isOpen = false;
+  }
+}());
--- a/site/src/app/components/dropdown/dropdown.html
+++ b/site/src/app/components/dropdown/dropdown.html
@@ -1,0 +1,8 @@
+<nav class="dropdown" ng-class="{ open: dropdown.isOpen }">
+  <div class="dropdown-current" ng-click="dropdown.isOpen = !dropdown.isOpen">
+    {{dropdown.selected}}
+  </div>
+  <div ng-click="dropdown.isOpen = false">
+    <ul class="menu" ng-transclude></ul>
+  </div>
+</nav>
--- a/site/src/app/components/language-switcher/language-switcher.directive.js
+++ b/site/src/app/components/language-switcher/language-switcher.directive.js
@@ -6,20 +6,14 @@
     .directive('languageSwitcher', languageSwitcher);
 
   /** @ngInject */
-  function languageSwitcher() {
+  function languageSwitcher(langs, manifest) {
     return {
       restrict: 'A',
       templateUrl: 'app/components/language-switcher/language-switcher.html',
-      controller: LanguageSwitcherCtrl,
-      controllerAs: 'switcher',
-      bindToController: true
+      link: function(scope) {
+        scope.langs = langs;
+        scope.modules = manifest.modules;
+      }
     };
-  }
-
-  /** @ngInject */
-  function LanguageSwitcherCtrl(langs) {
-    var switcher = this;
-
-    switcher.langs = langs;
   }
 }());
--- a/site/src/app/components/language-switcher/language-switcher.html
+++ b/site/src/app/components/language-switcher/language-switcher.html
@@ -30,25 +30,25 @@
     <h4 class="list-item--heading">External Resources</h4>
   </li>
   <li class="invisible-lg">
-    <a ng-href="https://github.com/GoogleCloudPlatform/gcloud-{{::lang}}" title="google-cloud on Github" class="skip-external-link">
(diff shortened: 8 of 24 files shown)


quartzmo:
@callmehiphop Fantastic, taking a look now!

/cc @blowmage

stephenplusplus:
Where do the directories `google-cloud` and `vision` go inside of the `gh-pages` branch? i.e. is it:

```sh
$ gc gh-pages
$ cd json
$ ls
 ├─── google-cloud
 |     └─── master
 |           ├─── index.json
 |           ├─── types.json
 |           ├─── toc.json
 |           └─── vision
 |                 └─── index.json
 └─── vision
       ├─── master
       |     ├─── index.json
       |     ├─── types.json
       |     └─── toc.json
       └─── v0.2.0
             ├─── index.json
             ├─── types.json
             └─── toc.json
```

callmehiphop:
@stephenplusplus yep, that's it exactly.

stephenplusplus:
Cool. Do we just leave all the stuff that's in there currently in place?

callmehiphop:
> Cool. Do we just leave all the stuff that's in there currently in place?

No.. basically everything that ends up in the current <kbd>json</kbd> folder would move to <kbd>json/google-cloud</kbd>. And for each module you would need to have a copy of <kbd>toc.json</kbd>, <kbd>types.json</kbd>, <kbd>overview.html</kbd> and any markdown files specific to that version.

quartzmo:
Any chance of a demo?

callmehiphop:
> Any chance of a demo?

Updated the PR overview with links to a couple different demos.

quartzmo:
@callmehiphop Are you still planning to have a package (module) select like [the one shown here](https://github.com/GoogleCloudPlatform/gcloud-common/issues/52#issuecomment-225016782)?

callmehiphop:
@quartzmo sure am, that's the easy part so I've been putting it off.

daspecster:
Would should the landing pages look like in this case? Should the "Read Docs" button go somewhere different or maybe change the layout so it's more like cloud.google.com's doc nav?

In this screenshot, instead of it being "Quickstarts"..."How To guides" etc...it could be "gcloud-compute"..."gcloud-datastore" etc...

screen shot 2016-07-29 at 5 25 37 pm

callmehiphop:
@daspecster that's a good point, I hadn't even really considered the homepage. Maybe put a mock together and open a new issue for all to discuss?

daspecster:
@callmehiphop, ok I'll see what I can come up with. I'm no designer though.

stephenplusplus:
@callmehiphop what's left to do here?

Can you give a step-by-step for what any lib that is modularized needs to do to enable the modularized site, preferably in a section in <kbd>site/README.md</kbd>.

And just a question, will module docs be able to link to other module docs? I.e.:

```js
/**
 * This is a Vision method that accepts a File as an argument.
 *
 * @param {module:storage/file} file - A File object.
 */
```

I think ideally it would link to the latest version of the Storage module's File docs. Is that tricky to implement, and if so, what would be the easiest way to handle that situation?

callmehiphop:
> what's left to do here?

I'll update the PR overview

> Can you give a step-by-step for what any lib that is modularized needs to do to enable the modularized site, preferably in a section in site/README.md.

Yeah, my plan was to do just that - I just wanted to get a few eyes on it to make sure there weren't any obvious red flags for anyone before I did.

> will module docs be able to link to other module docs?

It's definitely possible, but I would want to nail down the exact functionality we are expecting. I'm assuming that if we're browsing the giant package (`google-cloud`) we would not want it to link out to the individual package, whereas if they are looking at an individual package we would obviously have to link them to the separate module.

stephenplusplus:
> I'm assuming that if we're browsing the giant package (google-cloud) we would not want it to link out to the individual package

That's a good point. What you said sounds good to me-- if you're in a module, link to another module. If you're in the bundled package, link to the module inside of the bundle.

blowmage:
How do we format links to entities in a different package?

callmehiphop:
> How do we format links to entities in a different package?

This is where it gets tricky I think. Assuming you're in a standalone package (e.g. Datastore) if the app came across a link to the Storage module, I think we would just assume it wants the standalone package since there is no Storage content within the Datastore module. 

However, if you're looking at the main package that includes everything, I think we'll need a flag if we want to link out. My initial thought is to add a data attribute letting us know to link out:

```html
<a data-custom-type="storage/file" data-module-view="true">File</a>
```

Not sure how much of a pain that would be to implement though - and of course I'm open to suggestions!

stephenplusplus:
We could just write out a direct URL when we're doing something where automation doesn't need to be expected. The bundled package linking to a sub-module should be a rare case, right? Is there an obvious need for I'm not thinking of?

callmehiphop:
> Is there an obvious need I'm not thinking of?

My thought was basically if you are viewing the main package, you might want a line of text that says something like "You are viewing the main package, to view the Datastore package go here" at which point you link them to the individual module.

> We could just write out a direct URL when we're doing something where automation doesn't need to be expected. 

I think this has been brought up in the past and the general consensus was that we should avoid direct URLs to the docs where possible.

stephenplusplus:
Direct URLs in this case seem pretty reliable, since the URL end points should be consistent. Maybe we can implement the behavior we landed on a couple posts up and improve it if necessary later?

callmehiphop:
I've updated the demo with a working menu for the module switcher, there are still a couple small things to work out but otherwise this is essentially it. PTAL and let me know if anyone sees any potential roadblocks.

/cc @quartzmo @blowmage @stephenplusplus @jdpedrie @dwsupplee @daspecster

quartzmo:
@callmehiphop Now that GoogleCloudPlatform/gcloud-ruby#813 is open, I will attempt to deploy a gcloud-ruby demo this afternoon!

quartzmo:
@callmehiphop @stephenplusplus I have been running the demo locally and finally understand what you mean by giant/bundled package. Our expectation was for something different: A link to "BigQuery" would be a navigation to the individual service package, in the correct version for the current version of the bundle. As far as the build goes, instead of copying the json for bigquery into the directory for the bundle, it would only build the bigquery json to the directory for the individual service package. 

In the site, the navigation to the individual package BigQuery would be reflected in the displayed version. Currently, I see nothing that indicates which version of BigQuery is inside the bundle.

I'm not sure how exactly these cross-package links would work, but we need to solve that anyway, don't we? For example, to navigate to Storage from BigQuery?

callmehiphop:
> A link to "BigQuery" would be a navigation to the individual service package, in the correct version for the current version of the bundle.

I'd like to discuss this a bit more, I believe the new module dropdown would do something very similar to what you're describing - it would take you to the individual service package.

Are you also suggesting that when viewing the bundled documentation, clicking a reference to Storage should take you to the individual Storage service documentation as well? 

> As far as the build goes, instead of copying the json for bigquery into the directory for the bundle, it would only build the bigquery json to the directory for the individual service package.

This is still doable, we could just create a JSON file that captured the available services and versions within a bundle.

> In the site, the navigation to the individual package BigQuery would be reflected in the displayed version. Currently, I see nothing that indicates which version of BigQuery is inside the bundle.

Initially I didn't think this was necessary, from a user's perspective if they installed `google-cloud` v1.0.0, I assumed that they probably weren't going to care what the individual service version was. I'm not sure what the use case for this would be, but if you have one I'd gladly fit it in.

> I'm not sure how exactly these cross-package links would work, but we need to solve that anyway, don't we? For example, to navigate to Storage from BigQuery?

This part is definitely a bit tricky, my thoughts are essentially that:

* If a user is browsing the bundled package, we should not link to individual services packages since the bundled package is all inclusive.
* If the user is browsing an individual service package, any links for other services should direct the user to the other individual service packages. When linking to a different service package, we should always go to the latest release. My reasoning being that there is no guarantee that both packages are up to date on the users machine.

WDYT?

quartzmo:
@callmehiphop This seems to be kind of a "union vs confederation" question, and I have to admit that right now I'm swayed by your response and uncertain about which is better. It is significant however that @blowmage and I were both surprised by the choice to make the bundled package appear and feel seamless and unified. I guess we assumed that the point was to promote and emphasize the granularity of the separate packages. As far as the user experience goes, this would mainly just mean prominently displaying the correct individual package version, even when navigating from the bundle entry point. (To get the versions correct may require some metadata capturing the versions of the dependencies in the bundle?)

I'm limiting my observations to the user experience, but I think @blowmage has some thoughts about the value of not duplicating json between bundle and individual packages in the build process.

blowmage:
> Are you also suggesting that when viewing the bundled documentation, clicking a reference to Storage should take you to the individual Storage service documentation as well?

This is my opinion. Our umbrella package has no code in it, it only defines dependencies to the individual packages. I find it confusing that the umbrella/bundled package is presented as if the code is contained within it.

> Initially I didn't think this was necessary, from a user's perspective if they installed google-cloud v1.0.0, I assumed that they probably weren't going to care what the individual service version was. I'm not sure what the use case for this would be, but if you have one I'd gladly fit it in.

Here is a scenario: `google-cloud` v1.0.0 and `google-cloud-bigquery` v1.0.0 are released. The `google-cloud` package has a dependency on `google-cloud-bigquery` versions `>= 1.0.0` and `< 2.0.0`. User A installs both gems at `1.0.0`. Two months later we release `google-cloud-bigquery` version v1.1.0, which adds the `legacy_sql` experimental feature. The change is additive, and so it is a minor release. As such, google-cloud does not see an associated release. User B hears that the `legacy_sql` feature was added and installs `google-cloud` v1.0.0, which also installs the latest `google-cloud-bigquery` which is now v1.1.0.

What does the documentation show?

If the bundled docs show the latest, then User B is happy. But, how are users made aware that the `legacy_sql` feature is part of bigquery v1.1.0, and not v1.0.0? Also, User A can become confused because the documentation for the version of the library they are using changes on them without warning and documents a feature that is not present in the versions of the packages they are using.

If the bundled docs show the docs as `google-cloud` was when first released, at v1.0.0, then User B can become confused because the feature they heard was added is not present in the documentation.

Now multiply this by a dozen individual packages, all with their own independent release schedule, and this problem compounds exponentially. Hence my opinion that docs should belong to the versioned individual packages only.

callmehiphop:
>  I find it confusing that the umbrella/bundled package is presented as if the code is contained within it.

I can see where you're coming from with this, but my thinking is that the user isn't really going to care about where the actual code lives, but rather what code they have to work with when installing a particular package, whether that be an individual package or the umbrella. 

> User A can become confused because the documentation for the version of the library they are using changes on them without warning and documents a feature that is not present in the versions of the packages they are using.

This is a fair point, but I find it unsettling because now we're entering territory where the user has to be aware of our dependency versions in order to view the proper documentation. Which kind of defeats the point of the umbrella package. 

IMO this approach will make our docs more confusing and harder to navigate. I think instead we should revisit how we version the umbrella package. What are your feelings on bumping the minor of the umbrella when a minor version increment happens on an individual package?

quartzmo:
> the user has to be aware of our dependency versions in order to view the proper documentation

I'm trying to think of a mechanism that brings the user to the correct latest version per the actual package dependencies. Seems like it could work...

callmehiphop:
> I'm trying to think of a mechanism that brings the user to the correct latest version per the actual package dependencies. Seems like it could work...

I think we'll actually run into the same issue that @blowmage described earlier. We could send them to the latest version of the package dependencies, but that doesn't necessarily mean that's what is installed on their machine. At which point they could potentially notice documentation for features that are missing from their local package. Then the only way to navigate to correct documentation would be by examining the dependency version of what is installed or by re-installing altogether.

blowmage:
> We could send them to the latest version of the package dependencies, but that doesn't necessarily mean that's what is installed on their machine.

I think using the latest version of each package might work as long as each code entity (classes and methods) were properly labeled with the package and version it was from. We run into problems when we hide the package details, IMO.

callmehiphop:
> I think using the latest version of each package might work as long as each code entity (classes and methods) were properly labeled with the package and version it was from. We run into problems when we hide the package details, IMO.

This is a feature that is easy enough to add while still providing full documentation within the umbrella package. However, it doesn't change the fact that even if it's clearly labeled the user still has to dig into our package dependencies to verify they're looking at the correct documentation in the event that a minor update was made to an individual package.

IMO the correct way to go about fixing this is by re-visiting how we version the umbrella package.

blowmage:
> I think instead we should revisit how we version the umbrella package. What are your feelings on bumping the minor of the umbrella when a minor version increment happens on an individual package?

I think that would negate many of the advantages of shipping individual packages. It seems like a pretty radical solution for what is essentially a presentation issue on the documentation site.

stephenplusplus:
I think the docs just brought the question to light. I was always under the assumption if a sub-package shipped a feature, the user wouldn't need to do anything, they'd just get it. And even if we bump the umbrella minor when a sub-package gets a feature, that's still true.

I'm sorry that this will be Node-biased, I'm just not sure how it will work in other languages: a Node project depending on the umbrella package will (by default) install any minor bumps automatically when the package is re-installed/updated. The default tag from an `npm install google-cloud` will be something like `^1.0.0`, which is `>=1.0.0 && <2.0.0`.

So, in terms of using semver to communicate changes to users... when a subpackage gets a feature, our umbrella package gets a feature... which is a feature, so we should tag it as one. That's where I'm at currently.

blowmage:
I think your statement reinforces the scenario I outlined earlier. If you install a previous version of the `google-cloud` umbrella package, both NPM and RubyGems will return the most up-to-date versions of the individual packages, because of the permissive dependencies declared in the umbrella package. User A's documentation changes without warning, or User B doesn't get the documentation they expect.

Bumping the umbrella package version when a sub-package gets a feature will only help if the umbrella package declares exact dependencies on the individual packages. For example, `google-cloud` v1.0.3 depends on `google-cloud-bigquery` v1.1.0. Not `>= 1.1.0`, but the exact version `= 1.1.0`. This would negate all of the advantages of shipping individual packages and we might as well stay with the current monolithic package.

Again, the problem as I see it is not how we version or declare dependencies, but in how the  "bundled" package documentation obscures the fact that the code constructs live in a specific version of the individual package. If the package and version was displayed as part of documentation presentation then I think we would avoid these problems. I would rather see the angular application get much better at presenting the documentation across individual packages than duplicating documentation in the "bundled" package, or changing how we declare package dependencies and release process.

stephenplusplus:
Doesn't that go away if we declare versions to only include bug fixes? So ~1.1.0 which is >1.1.0 && <1.2.0. Modularization features are not negated, they are the same, with less surprises.

quartzmo:
@stephenplusplus @blowmage Although relevant here for sure, this discussion of dependency versioning seems like it should move to a higher-level gcloud-common issue. It seems like an important decision.

stephenplusplus:
The plan for now is to make the docs work for the default semver behavior of marking breaking changes, features, and bug fixes, which is what we are going to use in gcloud-node land (re: above versioning discussion).

On the umbrella docs linking to module docs, we should move forward with the umbrella site including JSON from the submodules, so that the docs are welcoming. We can continue to improve the docs from collective input (including user feedback), but to get v1 module support out the door, lets ship what we've got.

I'm glad this led to the versioning discussion, which is ultimately more important than docs, since we can always change how we construct our website. If you think we all should be using semver differently, lets indeed take that to a new issue.

quartzmo:
Regarding the docs, sounds good. I am still working on producing a demo for gcloud-ruby using this PR.

quartzmo:
#### Demo for gcloud-ruby

[Source](https://github.com/quartzmo/gcloud-ruby/tree/gh-pages) 

[Demo](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.12.0/google/cloud)

Note that the "bundled" package contains just BigQuery and Datastore for now.

quartzmo:
@callmehiphop While working on the json, I noticed that having bad values in some parts of the path can cause the app (and browser tab) to hang indefinitely. Below is an example. Please see if it reproduces for you in my gh-pages deployment (this path hangs for me):

[/#/docs/google-cloud-bad-path/v0.12.0/google/cloud](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud-bad-path/v0.12.0/google/cloud)

blowmage:
I've outlined some thoughts on the angular documentation site on GoogleCloudPlatform/gcloud-common#159. PTAL. Thanks!

callmehiphop:
@quartzmo I made some updates to the routing logic, if you wouldn't mind taking a look it would be much appreciated :)

@stephenplusplus I think this is wrapped up if you don't mind taking a look, I've also done a push to the modular demo link

quartzmo:
@callmehiphop Thanks, I'll try it out and update my demo!

quartzmo:
@callmehiphop I updated my demo to the build from your demo:

```sh
commit 7cc49d704b457aab731cfea0e778c7e5a2fe8623
Author: Chris Smith <quartzmo@gmail.com>
Date:   Tue Aug 9 16:11:14 2016 -0600

    Update demo with gcloud-common 98df4ec

 index.html             | 4 ++--
 src/app-045735d99f.css | 3 +++
 src/app-7077940c93.css | 3 ---
 src/app-b538ba9bc6.js  | 1 -
 src/app-cf72eae2a8.js  | 1 -
 src/app-e6fb261ef2.js  | 1 -
 src/app-e791332cd7.js  | 1 +
 src/app-f71ddd10e9.css | 3 ---
 8 files changed, 6 insertions(+), 11 deletions(-)
```

For correct routes, the app is working fine. Here is a good route that works:

https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.12.0/google/cloud

Unfortunately, certain bad paths still cause my browser to hang in what appears to be an infinite redirect. Here is one:

https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud-bad-path/v0.12.0/google/cloud

Can you confirm that you can reproduce the hang if you click the link above? And if so, do you see anything in my [JSON source](https://github.com/quartzmo/gcloud-ruby/tree/gh-pages) that stands out? (Although it seems that for easier development and debugging, never hanging would be ideal.)

callmehiphop:
@quartzmo I can confirm! Made a new commit, PTAL!

quartzmo:
Trying it out now!

quartzmo:
@callmehiphop Seems to be working great! If you update your demo, I'll copy the new files to mine.

quartzmo:
@callmehiphop I am almost done preparing a new demo that includes all of the gcloud-ruby docs.

I just noticed a potential usability issue: Since the module switcher is not scrollable, some items may be hidden if the list is longer than the display area. See screenshot of entire browser window below:

screen shot 2016-08-10 at 4 18 52 pm

quartzmo:
The gcloud-ruby [demo](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/master/google/cloud) now includes all services, both as individual packages and a bundled package. Only master branch is included for now.

callmehiphop:
@quartzmo good catch on the CSS bug, I just pushed another commit that should resolve that. I updated my demo/files as well.

quartzmo:
@callmehiphop I updated the gcloud-ruby demo with the latest build. I also added `home.html` and bundled docs for our current release, so that the demo works from the root path:

https://quartzmo.github.io/gcloud-ruby/

PTAL and let me know if you can find anything wrong with the app or the JSON; at this point it should be production-ready, I think.

/cc @stephenplusplus @blowmage @jgeewax

quartzmo:
@callmehiphop Does this change provide backwards compatibility for existing JSON? I am unable to navigate to older versions. My `manifest.json` looks like this:

```js
{
  "lang": "ruby",
  "friendlyLang": "Ruby",
  "moduleName": "gcloud-ruby",
  "markdown": "ruby",
  "versions": [
    "v0.9.0"
  ],
  "titleDelimiter": "::",
  "defaultModule": "google-cloud",
  "content": "json",
  "home": "home.html",
  "package": {
    "title": "RubyGems",
    "href": "http://rubygems.org/gems/gcloud"
  },
  "modules": [
    {
      "id": "google-cloud",
      "name": "google-cloud",
      "defaultService": "google/cloud",
      "versions": [
        "v0.12.2",
        "master"
      ]
    },
    {
      "id": "google-cloud-bigquery",
      "name": "google-cloud-bigquery",
      "defaultService": "google/cloud/bigquery",
      "versions": [
        "master"
      ]
    },
    ...
```

Contents of `json` directory:

```
google-cloud
google-cloud-bigquery
google-cloud-core
google-cloud-datastore
google-cloud-dns
google-cloud-language
google-cloud-logging
google-cloud-pubsub
google-cloud-resource_manager
google-cloud-speech
google-cloud-storage
google-cloud-translate
google-cloud-vision
home.html
v0.9.0  
```

callmehiphop:
@quartzmo It doesn't support both the modular AND regular view at the same time. For node we just moved all the old package versions to the new [umbrella version list](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/docs/manifest.json#L9) and then moved the old JSON to the [umbrella json folder](https://github.com/GoogleCloudPlatform/gcloud-node/tree/gh-pages/json/google-cloud).

quartzmo:
@callmehiphop Thanks for the crucial tip! I'll try doing what you did.

quartzmo:
@callmehiphop 

There is a small issue with older versions. (So that you can see, I've deployed to my demo the latest site build from `master` and added `v0.9.0` json to the umbrella folder.)

I can successfully navigate directly to the older docs: [/#/docs/google-cloud/v0.9.0/gcloud](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.9.0/gcloud)

But because I need to configure `defaultService` as `google/cloud` for the latest version, and this configuration is scoped to all versions, using the switcher to change to `v0.9.0` brings me to [#/docs/google-cloud/v0.9.0/google/cloud](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.9.0/google/cloud), which is redirected to the latest version.

quartzmo:
In our older docs, the `defaultService` is `gcloud`.

callmehiphop:
@quartzmo hmm that is quite the problem. You could always make a separate package in the manifest called something like `gcloud - DEPRECATED` rather than lumping the two together?

quartzmo:
@callmehiphop I'll try making a `gcloud` "module".

quartzmo:
@callmehiphop Thanks for the quick responses, I finally understand how to publish the older docs! Works great: https://quartzmo.github.io/gcloud-ruby/#/docs/gcloud/v0.9.0/gcloud

PR #165, opened 2016-08-18
Expose description on method return

jdpedrie:
Add the return description, if set, after the type. Description is rendered on a different line because the markdown parser wraps it in a paragraph tag.

screen shot 2016-08-18 at 3 46 23 pm

Without description, no change:

screen shot 2016-08-18 at 3 48 44 pm

(Those screenshots are from different methods)

### DIFF
--- a/site/src/app/docs/docs.service.js
+++ b/site/src/app/docs/docs.service.js
@@ -51,7 +51,10 @@
     }
 
     function trustReturn(returnValue) {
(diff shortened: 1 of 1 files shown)


callmehiphop:
Can you give a quick demo of what this looks like (with & without a description). Screenshot is fine :)

jdpedrie:
@callmehiphip I updated the first comment

PR #167, opened 2016-08-23
Fix navigation expansion by not splitting serviceId.

daspecster:
Fixes  side navigation always being expanded and not collapsing.

### DIFF
--- a/site/src/app/docs/docs.controller.js
+++ b/site/src/app/docs/docs.controller.js
@@ -38,7 +38,6 @@
     }
 
     function isActive(serviceId) {
-      serviceId = serviceId.split('/')[0];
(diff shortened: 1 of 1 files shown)


callmehiphop:
Thanks @daspecster!

dwsupplee:
The reason this was here in the first place was to fix an issue in PHP site where nested menu items were not being highlighted or expanded when clicked on.

Maybe this is something we need to think on a little bit more so that we can solve for both python and php.

daspecster:
@dwsupplee can you give an example ID or a JSON file from https://github.com/GoogleCloudPlatform/google-cloud-php/tree/gh-pages?

PR #168, opened 2016-08-23
Add gcloud-python to docsite updates.

daspecster:
Add `gcloud-python` to receive auto doc site updates/deployments.

### DIFF
--- a/deploy-docs.sh
+++ b/deploy-docs.sh
@@ -71,3 +71,4 @@ function deploy_docs {
 }
 
 deploy_docs "googlecloudplatform/gcloud-node"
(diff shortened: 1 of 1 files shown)


deploy-docs.sh
```diff
@@ -71,3 +71,4 @@ function deploy_docs {
 }
 
 deploy_docs "googlecloudplatform/gcloud-node"
+deploy_docs "googlecloudplatform/gcloud-python"
```
stephenplusplus: @callmehiphop do we need to update anything here to correct the URL for the repo?

daspecster: @stephenplusplus see #170.

PR #169, opened 2016-08-23
docs: adding support for partial service ids

callmehiphop:
This adds support for full & partial service ID matching when deciding on whether or not to expand/collapse a navigation item.

### DIFF
--- a/site/src/app/docs/docs.controller.js
+++ b/site/src/app/docs/docs.controller.js
@@ -38,7 +38,13 @@
     }
 
     function isActive(serviceId) {
-      return !!($state.params.serviceId || '').match(serviceId);
+      var serviceIdParam = $state.params.serviceId || '';
+      var isCompleteMatch = !!serviceIdParam.match(serviceId);
+
+      var partialServiceId = serviceId.split('/')[0];
+      var isPartialMatch = !!serviceIdParam.match(partialServiceId);
+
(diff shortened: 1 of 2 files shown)


dwsupplee:
Just tested this out and it works great for the PHP docs, thanks! :)

callmehiphop:
@dwsupplee glad to hear it :)

@daspecster did you have a chance to test this out?

callmehiphop:
Nevermind, totally doesn't work for Python

daspecster:
Seems to break it again :(

screen shot 2016-08-23 at 12 09 03 pm

site/src/app/docs/docs.controller.js
```diff
@@ -38,7 +38,14 @@
     }
 
     function isActive(serviceId) {
-      return !!($state.params.serviceId || '').match(serviceId);
+      var serviceIdParam = $state.params.serviceId || '';
+
+      if (!manifest.matchPartialServiceId) {
```
callmehiphop: @dwsupplee I had to take a peak at your JSON to remember why we needed to split the params and everything came back to me. I feel that making this configurable is the least destructive path. WDYT?

dwsupplee: LGTM

daspecster: Works for python!

PR #170, opened 2016-08-23
gcloud-node -> google-cloud-node in deploy-docs.sh.

daspecster:
@callmehiphop per us talking about this earlier and @stephenplusplus [mentioning](https://github.com/GoogleCloudPlatform/gcloud-common/pull/168#discussion_r75883190) it.

Github has repo redirects appear to stay in place until you make a new repo of the same name as the one you renamed. So in theory it looks like the redirect would last a long time.

### DIFF
--- a/deploy-docs.sh
+++ b/deploy-docs.sh
@@ -76,5 +76,5 @@ function deploy_docs {
   rm -rf ../.git/modules/gh-pages
 }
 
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thanks!

callmehiphop:
Thanks @daspecster!

PR #182, opened 2016-09-06
site: add 404 page and setup unit tests

callmehiphop:
This PR aims to accomplish a few things

* Include necessary scripts to run unit tests
* Include a 404 page
* Update redirect logic to send the user to said 404 page
* Run unit tests on PRs

===

[demo](http://callmehiphop.github.io/gcloud-node/#/docs/wat)


callmehiphop:
@quartzmo @dwsupplee @daspecster If you guys could test these changes out, it would be greatly appreciated :)

stephenplusplus:
Tested with a URL in the old format:

http://callmehiphop.github.io/gcloud-node/#/docs/v0.36.0/bigquery/dataset?method=createTable

Got redirected to the latest release's home page: http://callmehiphop.github.io/gcloud-node/#/docs/google-cloud/0.39.0/google-cloud

Ideally that would be redirected to the correct place: http://callmehiphop.github.io/gcloud-node/#/docs/google-cloud/0.36.0/bigquery/dataset?method=createTable

If that's not doable, a 404 would probably be better :\

callmehiphop:
@stephenplusplus ahh good catch. Looks like the cause is because we no longer have the *v* in the version name. I'll make a quick update :)

quartzmo:
@stephenplusplus Should we stop using the "v" in the version number?

@callmehiphop I'll test it this afternoon.

callmehiphop:
@stephenplusplus so, this is probably more of a node specific issue, but it wasn't quite as simple to fix the version format change as I thought. I patched it by creating version folders in `google-cloud` for `v0.x.x` as well as keeping the `0.x.x` folders. Not the prettiest I know.. but the least painful.

quartzmo:
@callmehiphop I pulled and built the app, deployed to my fork gh-pages with our latest JSON. I can't see anything wrong. The 404 seems to work: https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.20.0/not-found

daspecster:
@callmehiphop, I'll message you on gchat.

stephenplusplus:
> @stephenplusplus Should we stop using the "v" in the version number?

I think that's more of a @callmehiphop question.

> I patched it by creating version folders in google-cloud for v0.x.x as well as keeping the 0.x.x folders.

Looking in from the outside, having our app do a `/^v(\d)/.replace('$1')` seems like a quick fix.

callmehiphop:
> Looking in from the outside, having our app do a /^v(\d)/.replace('$1') seems like a quick fix.

My hesitation there is that for other apps who do use `v` we'd be unable to locate the JSON.

stephenplusplus:
How about if we get a 404, append the `v` and carry on?

callmehiphop:
@stephenplusplus @quartzmo I've made a few small updates to do the following

* remove `v` present only in the event of a state change error
* disregard trailing slashes - angular-ui-router interprets it as a parameter.

PTAL!

stephenplusplus:
@callmehiphop can we do it without duplicated folders?

callmehiphop:
@stephenplusplus sure can!

Issue #137, opened 2016-05-20
Can we add link anchors to headings in MD pages?


jgeewax:
I want to link to a specific heading on the overview page on gcloud-ruby, and can't unless I open up the inspector and find the ID for the heading and then do it manually.

Can we make headings into hoverable links so I can right click and do "copy link location" to share a specific section?

Ie, https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud/datastore#paginating-records (I had to create that myself)

/cc @quartzmo @callmehiphop

stephenplusplus:
Duplicate: https://github.com/GoogleCloudPlatform/gcloud-common/issues/136 -- TLDR, we may have an ingenious fix. We just need a volunteer to try it.

callmehiphop:
@stephenplusplus I think the functionality between that issue and this request are slightly different. While #136 would allow us to use the `#` for deep linking, I believe this issue is asking for heading tags in markdown files to generate a deep link url when clicked (regardless of whether we use the `#` or a query string).

stephenplusplus:
Okay, pretending this never happened.

callmehiphop:
Looks like this isn't an option of showdown, not from what I can tell at least. @stephenplusplus did we have deep linking capabilities into the markdown guides on the old site? Either way, I think we could just find/write an extension to do it.

jgeewax:
Bump :)

Trying to link again to specific sections (and examples) and I can't. Which makes me sad.

![](http://www.mememaker.net/static/images/memes/3675821.jpg)

jgeewax:
@callmehiphop : https://github.com/showdownjs/showdown/compare/master...jgeewax:patch-1

Thoughts?

callmehiphop:
@jgeewax I think the issue here is that the `#` is already spoken for by Angular.

We need to either:

1. Resolve #136 
2. Add a custom extension to add deeplinking

daspecster:
@callmehiphop what do you think of that `404.html` idea?

callmehiphop:
@daspecster I haven't looked into it very deeply (yet), but it seems pretty clever and the rest of the group seems to be in favor of it.

jgeewax:
As much as a I love the ingenuity.... it feels like one of those things that will bite us in the ass later, and then we'll be too far committed to it to fix things properly. 

Can we explore making the header ID a sub-resource? ie,

```html
[...]

Issue #40, opened 2015-11-30
No anchor links to pin to subheadings?


stephenplusplus:
_From @jgeewax on September 29, 2015 15:50_

I wanted to send a reference to the "With gcloud-node" section on http://googlecloudplatform.github.io/gcloud-node/#/authentication but it looks like there's no paragraph anchor link...

Any chance we can add one of those?

_Copied from original issue: GoogleCloudPlatform/gcloud-node#899_

stephenplusplus:
_From @callmehiphop on September 30, 2015 5:57_

It doesn't look like any of the markdown content has anchor tags.. was it always like this? Regardless, I'm not sure if showdown offers this as an option so we may have to write an extension.

stephenplusplus:
_From @jgeewax on September 30, 2015 14:42_

I'm not sure if we ever had this working -- I suspect not... however maybe h2 and higher should have anchors? Or all headings?

Not the most important thing, but definitely something in the "nice to have" category.

quartzmo:
The title of this issue should be generalized to address all headings, or else should be closed as a duplicate of later issue #137.

PR #189, opened 2016-09-22
site: create anchor directive

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/913

### DIFF
--- a/site/src/app/components/deep-link/anchor.directive.js
+++ b/site/src/app/components/deep-link/anchor.directive.js
@@ -1,0 +1,21 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('gcloud')
+    .directive('anchor', anchorDirective);
+
+  /** @ngInject */
+  function anchorDirective($location) {
+    return {
+      restrict: 'A',
+      link: function(scope, elem, attrs) {
+        var anchor = attrs.anchor;
+        var href = '#' + $location.path() + '?section=' + anchor;
+
+        elem.attr('href', href);
+      }
+    };
+  }
+
+}());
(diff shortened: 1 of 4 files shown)


callmehiphop:
@quartzmo PTAL!

quartzmo:
Nice having the spec to show it working. I'll pull the branch and try it now, but pretty clear that it works!

quartzmo:
@callmehiphop The links work great internally, but external URLs do not bring me to the anchor, just the top of the page. Local example:

http://localhost:3000/#/docs/google-cloud-logging/master/google/cloud/logging?section=#exporting-log-entries

Should we worry about this?

callmehiphop:
@quartzmo I think it should be http://localhost:3000/#/docs/google-cloud-logging/master/google/cloud/logging?section=exporting-log-entries

quartzmo:
Ok, that's good:`?section=exporting-log-entries` does work, the problem now is just that when I click the internal anchor link, the browser bar displays `?section=#exporting-log-entries`. If I were trying to share the location, I would naturally copy and paste that, resulting in a broken link.

callmehiphop:
So just to clarify, your HTML resembles the following

```html
<a data-anchor="#exporting-log-entries">Link Text</a>
```

?

callmehiphop:
@quartzmo I added a new commit + test to ignore hash characters. PTAL

quartzmo:
Aha, yes! (I just manually changed the `href` to `data-anchor`). I removed the `#` and it's good. But nice that you updated the code to prevent the issue...

quartzmo:
I tested the update, works great for me now, with or without a hash in the value.

I'm ready to merge after build is green. Should my merge commit close #137? What about #40?

callmehiphop:
I think we can safely close #137, but I would leave #40 open because we don't generate the clickable anchors for the user yet.

quartzmo:
Thanks @callmehiphop!

callmehiphop:
@quartzmo np!!

PR #192, opened 2016-09-28
site: add Go

jba:
Add Go to lang config.

Add images for the dropdown and for the red doc button.


site/src/app/index.constants.js
```diff
@@ -29,6 +29,10 @@
       friendly: 'Ruby',
       key: 'ruby',
       repo: 'google-cloud-ruby'
+    }, {
```
stephenplusplus: We've been keeping this alpha-sorted, so feel free to bump up to above Java ☕ .

jba: Done.

callmehiphop:
@jba I noticed that the Go icon is in color, which none of the other icons are. Any chance we could get a white icon instead?

jba:
@callmehiphop I'm pretty confused by this. I spent a few hours trying to figure out what was going on. In the end I think it may have just been an old colored version of the file and a cache somewhere.

If you click the View button on the file on the this PR's File changed page, you'll see the svg rendered with no color.  You can also see the same file rendered on the [go site](https://googlecloudplatform.github.io/google-cloud-go) by clicking the dropdown, where it looks fine.

callmehiphop:
@jba ahh, that makes sense.

Issue #207, opened 2016-11-27
GAPIC Docs Integration with Ruby/PHP/Node/Python


omaray:
Hello all,

As you know, we want to integrate the GAPIC documentation with the current angular app documentation for Ruby/PHP/Node and with ReadTheDocs for Python. This important not only for the APIs that will have GAPIC as the top most layer, but also for APIs that use GAPIC underneath the covers (some users might still want to access it).

To that end, here's a proposal on how to present the GAPIC client library reference. In the images below we use the Monitoring API as the example and in the Ruby language. But the presentation should adapt to each language's idiom.

In this first proposal we focus mainly on the API surface and on the underlying protocol buffer data types. What is missing is how to present the GAX classes and other types that are API-independent (like the [common data types](https://github.com/googleapis/googleapis/tree/master/google)). More info on that at the end and as next steps.

## API Version
To start, we will present the GAPIC client library reference of an API underneath the API's version. If we take Monitoring as an example, we observe that clicking on the Monitoring left nav link will show the major version to the user - in this case V3. This way we can get multiple versions living side-by-side (though we don't expect more too many at any given time).

![image](https://cloud.githubusercontent.com/assets/6609430/20652515/8dc5455c-b4af-11e6-9438-c88522de033a.png)

## API/Client Surface
Once the user clicks on the API Version, we then present to them the client classes and also the data types that come from the proto layer. Starting with the client surface, we observe below that we simply show the user the documentation for the main GAPIC client classes. What we currently have today for Ruby and Monitoring works well (see [GroupServiceApi](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.22.0/google/cloud/monitoring/v3/groupserviceapi) and [MetricServiceApi](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.22.0/google/cloud/monitoring/v3/metricserviceapi)).

![image](https://cloud.githubusercontent.com/assets/6609430/20652584/bebb8774-b4b0-11e6-9fc3-3d9fd49b3318.png)

The only thing missing from the client classes shown above are links to the non-native types that are referenced in the documentation (e.g. the types of the parameters or return values for proto classes).

## Data Types
[...]

Issue #85, opened 2016-02-08
Add API Reference


blowmage:
Assuming #78 is accepted, we suggest adding a new feature that will list all code constructs and methods in the version. The gcloud-ruby documentation does this currently. It has the [curated side navigation](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/), and an [API Reference](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/reference.html) with a full listing of all the costructs and methods. This is very useful for gcloud-ruby, which has many more code constructs than can fit in the side navigation.

stephenplusplus:
I think this relates to what I was asking here: https://github.com/GoogleCloudPlatform/gcloud-common/issues/78#issuecomment-181445609

Is a happy alternative to a new API reference page just placing the "extra" constructs and methods separate from the common case ones?:

```
API
> gcloud
> BigQuery
  > Dataset
  > Table
  > Job
  > API
    > ApiError
    > CopyJob
    > Data
    > Error
    > ExtractJob
    > ...
> Compute
> DNS
...
```

I think that-- or something like it-- gives the user a quicker way to first find the thing they're looking for, then go back and forth between it and others.

quartzmo:
@stephenplusplus I think what @blowmage calls an API Reference, I think of as a good ole-fashioned *index*. Our current [gh-pages API Reference](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/reference.html) does *not* provide a global listing of all public methods, but [our docs on rubydoc.info](http://www.rubydoc.info/gems/gcloud) provide this when you click `Methods` in the left nav. Pretty handy when you know a method name but can't remember where it is!

callmehiphop:
@quartzmo based on the [version gist](https://gist.github.com/blowmage/13be352f40ee7ac5ee48#file-version-json) I assumed this would be a new feature that would essentially mimic the functionality on rubydocs.

It's probably worth mentioning that we've been discussing similar functionality on a [gcloud-node issue](https://github.com/GoogleCloudPlatform/gcloud-node/issues/1084), so I think that @stephenplusplus was trying to start a conversation about how we can implement an API Reference in an intuitive fashion that works for everyone.

blowmage:
We need to preserve the behavior of our current navigation, which presents a curated view into the library. Adding new links in the navigation won't be consistent with the current approach. Therefore I don't think a new API section in the navigation is a good solution.

[...]

PR #213, opened 2016-12-23
site: Add level of nesting to left nav

### LINKED ISSUE #209: Site left nav does not support adequate nesting
To deliver the design proposed in #207, the Angular site must be updated (per [this comment](https://github.com/GoogleCloudPlatform/gcloud-common/issues/207#issuecomment-265888784) by @callmehiphop) to support nesting of items in the left nav.

@callmehiphop Is this something you can do, or should I give it a try?

quartzmo:
[Demo](https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1)

In order to support the design in #207, this PR adds an extra level of nesting support in the left nav. I can produce a demo if needed. Here is a screenshot from my local app:

protobuf-data-types-nav

[closes #209]

/cc @omaray @blowmage @swcloud

### DIFF
--- a/site/src/app/styles/_main.scss
+++ b/site/src/app/styles/_main.scss
@@ -1151,6 +1151,10 @@ ul {
     padding-left: 4em;
 }
 
+.side-nav .sub-sections-2 a {
+    padding-left: 5.5em;
+}
+
 .side-nav .external-links {
     margin-top: 2em;
 }
--- a/site/src/app/docs/docs.controller.js
+++ b/site/src/app/docs/docs.controller.js
@@ -38,15 +38,27 @@
       return $state.go($state.current.name, { version: version });
     }
 
-    function isActive(serviceId) {
+    function isActive(service) {
+      var serviceId = service.type;
       var serviceIdParam = $state.params.serviceId || '';
 
-      if (!manifest.matchPartialServiceId) {
-        return !!serviceIdParam.match(serviceId);
+      // Match the first element in the serviceId. E.g., nav parent
+      // `datastore/client` matches `datastore` extracted from nav child
+      // `datastore/query`.
+      if (manifest.matchPartialServiceId) {
+        var partialServiceId = serviceId.split('/')[0];
+        return !!serviceIdParam.match(partialServiceId);
+
+      // Match the downcase service title to any part of the serviceIdParam,
+      // E.g., both nav parent `google/cloud/datastore` and nav child
+      // `google/datastore/v1` match the title `datastore`.
+      } else if (manifest.matchServiceTitle)  {
+        var parts = serviceIdParam.split('/');
+        return parts.indexOf(service.title.toLowerCase()) >= 0;
       }
 
-      var partialServiceId = serviceId.split('/')[0];
-      return !!serviceIdParam.match(partialServiceId);
+      // Strictly match. E.g., `datastore/query` matches `datastore`.
+      return !!serviceIdParam.match(serviceId);
     }
 
     function getGuideUrl(page) {
--- a/site/src/app/docs/docs.html
+++ b/site/src/app/docs/docs.html
@@ -42,9 +42,14 @@
       </li>
       <li ng-repeat="service in docs.services">
         <a side-nav-link="/{{service.type}}">{{service.title || service.type}}</a>
-        <ul class="sub-sections" ng-if="service.nav && docs.isActive(service.type)">
+        <ul class="sub-sections" ng-if="service.nav && docs.isActive(service)">
           <li ng-repeat="page in service.nav">
             <a side-nav-link="/{{page.type}}">{{page.title || page.type}}</a>
+            <ul class="sub-sections-2" ng-if="page.nav && docs.isActive(page)">
+              <li ng-repeat="page_2 in page.nav">
+                <a side-nav-link="/{{page_2.type}}">{{page_2.title || page_2.type}}</a>
+              </li>
+            </ul>
           </li>
         </ul>
       </li>


callmehiphop:
@quartzmo I think it looks good. I have a couple of questions:

1. Do you have to click on `V1` to expand the nested nav? Or does it auto-expand by selecting Datastore?

2. Should we be worried about text overflow here? I'm not sure where we'd see a super long name, but what happens when we do?

quartzmo:
You do need to click the nested item to expand it, and focus at every level works as it should.

As for text not fitting, let me try that out, and post a screenshot or a demo.

quartzmo:
@callmehiphop I can confirm that text overflow will be an issue:

text-overflow

If you have any ideas for a remedy, please let me know.

quartzmo:
Meanwhile, I've discovered another, more pressing problem that will require an update to this PR:

The nav focus does not work for namespace paths that are not under the `service.type`. For example, if the `service.type` is `google/cloud/datastore` and the id for the nested item is `google/datastore/v1/allocateidsresponse` (note that `cloud/` is missing), the focus does not work and the nav does not expand. I'm working on a fix.

quartzmo:
@callmehiphop I have update the commit for this PR with additional logic for nav focus that is accessed by a new flag in `manifest.json`: `matchServiceTitle`. If `matchServiceTitle` is present, the nav will be shown as active if the service title is present anywhere in the path. This solves a problem in google-cloud-ruby in which lower-level gapic types use the namespace path `google/datastore`, while the service id is `google/cloud/datastore`. This change should not affect usage without the new flag.

PTAL at the [demo](https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1).

quartzmo:
@callmehiphop @stephenplusplus Any chance of merging this today? I'm waiting on it to publish a number of fixes to our gh-pages docs, which are currently broken.

@swcloud How does the  [demo](https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1) look to you? OK?

site/src/app/styles/_main.scss
```diff
@@ -1151,6 +1151,10 @@ ul {
     padding-left: 4em;
 }
 
+.side-nav .sub-sections-2 a {
+    padding-left: 6em;
```
stephenplusplus: I think each nested row goes in increments of 1.5em, so this would match the nesting with a value of 5.5em.

quartzmo: Oh, that sounds good, and will perhaps keep titles from overflowing the box. I'll update. Thanks!

stephenplusplus:
The only thing I'm troubled by is not having a sidebar link to the sub-pages, like https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1/allocateidsresponse. I know we don't want to get into an endless nesting pattern, so maybe we can flatten them out and at least have an index of sorts?

quartzmo:
@stephenplusplus This design issue has been discussed on #207, with the [current idea](https://github.com/GoogleCloudPlatform/gcloud-common/issues/207#issuecomment-269647414) being to list these types in [a top-level "page"](https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1).

A bit of background on the design of the left nav for google-cloud-ruby: The intention for the docs experience since as long as I can remember has been a "curated" onboarding experience. We have many types in Ruby, and the important "entry" point are likely to get lost in list if we put them all in the left nav. In fact, even as I write this, a quick look at #85 reveals that @blowmage has already explained our situation:

> we suggest adding a new feature that will list all code constructs and methods in the version. The gcloud-ruby documentation does this currently. It has the curated side navigation, and an API Reference with a full listing of all the costructs and methods. This is very useful for gcloud-ruby, which has many more code constructs than can fit in the side navigation.

quartzmo:
@stephenplusplus I noticed that in [the types index table in this page](https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1) there is no cell padding, or it doesn't seem to work. Can I do something to fix this in this PR that does not break table design elsewhere (params lists)?

stephenplusplus:
I think those tables just need to have a class name `table` applied to them:

screen shot 2017-01-03 at 4 33 10 pm

On the nesting discussion, we can keep that discussion somewhere else, that's fine. Just to respond here to the points you brought up here:

> The intention for the docs experience since as long as I can remember has been a "curated" onboarding experience.

Yes, and that fits well with the handcrafted APIs. For the generated APIs, we don't want to hand-select what goes in the sidebar to show users the most important classes. Instead, everything that has its own page should go there so the user can decide-- in the quickest way possible-- what is important to them.

> We have many types in Ruby, and the important "entry" point are likely to get lost in list if we put them all in the left nav.

This is for the handcrafted or the generated? I think WhateverClient should be the entry point, bold. Under it (not even indented, necessarily), should be things likethat have their own unique address, like https://quartzmo.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/datastore/v1/allocateidsresponse.

> we suggest adding a new feature that will list all code constructs and methods in the version. The gcloud-ruby documentation does this currently. It has the curated side navigation, and an API Reference with a full listing of all the costructs and methods. This is very useful for gcloud-ruby, which has many more code constructs than can fit in the side navigation.

Yes, that's like a site map, which is a very good idea, especially for the handcrafted APIs. That, and a more helpful sidebar don't need to compete with each other-- they will both be helpful for different uses.

My view is that if we have a unique page, the user should be able to get to it without remembering how they stumbled on it. It's a usability pattern that we should follow.

quartzmo:
@stephenplusplus Good point to make a distinction between handcrafted and generated! I will follow your suggestion, if no one is opposed, and add all types to the left nav in my upcoming google-cloud-ruby PR for gh-pages. This PR is not affected either way, correct?

stephenplusplus:
Sounds good to me! It might need a CSS tweak then, but we can get to that when the time comes.

This LGTM, thanks for being patient with us for the review. If @callmehiphop doesn't have any notes, feel free to self-merge when ready!

swcloud:
I have a few comments here:

1. When you click on 'V1', it gives you an empty page. That's a bit strange. Can we do this instead: when you click on 'V1', the only thing it does is to show its sub items and it does not change the content of the right panel?

2. The table of data types has no borders.

3. When you click on one data type in the table, it changes the right panel to a separate page. I think the design is that 'Data Types' corresponds to a single page which contains all the content of data types. When a data type is clicked, we just jump to the segment of the page for the data type.

2 and 3 do not seem to comply with the design in #207.

quartzmo:
@swcloud Thanks for looking. Here are some answers:

1. The page is empty because `Google::Cloud::Datastore::V1` is undocumented. This is an opportunity to provide an overview of the GAPIC API. If you can point me to some content to include here, I will add it to the google-cloud-ruby docs PR I am working on.

2. The source of the table is a doc comment in `/lib/google/cloud/datastore/v1/doc/google/datastore/v1/datastore.rb`. Although it may be possible to use HTML in the comment, markdown is much more compact and readable. Unfortunately, however, I have not been able to specify a CSS class via Markdown -> YARD.

```ruby
module Google
  module Datastore
    ##
    # The `Google::Datastore::V1` module provides the following types:
    #
    # Class | Description
    # ----- | -----------
    # {AllocateIdsRequest} | The request for Datastore::AllocateIds.
    # {AllocateIdsResponse} | The response for Datastore::AllocateIds.
    ...
    # {RunQueryResponse} | The response for Datastore::RunQuery.
    # {Value} | Holds any of the supported value types and associated metadata.
    #
    module V1
```

3. A single page containing the docs for all of these types would be extremely long, so we are proposing the new design in the demo, as [discussed here](https://github.com/GoogleCloudPlatform/gcloud-common/issues/207#issuecomment-269647414).

quartzmo:
@swcloud Regarding 2, I will follow up with a PR for the google-cloud-ruby `gcloud-jsondoc` branch which adds the CSS `table` class to tables in JSON `description` fields. Then we will have this:

table-table

PR #227, opened 2017-07-25
site: Update service.html to always show the available methods section

quartzmo:
Always show available methods section. This will reveal the section in the many classes of google-cloud-ruby where the docs for the constructor are private. It should not result in a regression for other platforms, as long as constructors are public and listed first.

[fixes GoogleCloudPlatform/google-cloud-ruby#1593]


callmehiphop:
I just merged, hope that's ok :)

PR #232, opened 2017-11-07
firestore cross-language tests

jba:
These are tests for Firestore clients in all languages. The tests are expressed in proto format.


testing/firestore/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+- `proto`: the protobuffers defining the test format.
+
+- `testdata`: the tests.
+   - `*.textproto`: a single test in text proto format.
+   - `tests.binprotos`: all the tests in a single file. The protos are in binary
+      format, each preceded by its length encoded as a varint.
+
+Do not modify the tests here. Regenerate them using the Go program at
+`cloud.google.com/go/firestore/cmd/generate-firestore-tests`.
```
jskeet: Is that expected to be visible?

jba: That's a go import path. The file lives [here](https://code.googlesource.com/gocloud/+/master/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go). 

I've debated whether the program should also live here, instead of in the Go repo. WDYT?

jskeet: I definitely think it's a bit confusing as it is. It may be obviously a Go import path to Go users, but it's not obvious to others IMO. I think it would be better here, personally - I can imagine wanting to make a change to a test and having to work over three repos at the moment.

PR #234, opened 2017-11-10
testing/firestore: give test files names

jba:
Add a string suffix to each textproto file instead of a number.
That way when we add or remove tests, filenames of unchanged
tests will be stable.


jskeet:
That also helps a lot in terms of finding the relevant text proto when you're working from binaries - thanks!

jskeet:
LGTM - I haven't checked that each rename is right for the text involved; it's the principle that looks good. Let me know if you want me to check more thoroughly.

jba:
That's fine! No need for that level of checking.

PR #235, opened 2017-11-11
testing/firestore: improve test descriptions

jba:
Prepend the method being tested to the description.

Also, fix the path to the generator in the top comment.


schmidt-sebastian:
Would it be possible to create a different category for merge with fields paths (it might even make sense to have three categories: `set`, `set-merge-all`, `set-merge-paths`?). This would allow me to just ignore all tests that start with `set-merge-paths`.

Otherwise, this looks good to me and should help me ignore fewer tests.

jba:
I amended the commit: if a set dest has a merge with paths, its description begins "set-merge:".

PR #238, opened 2017-11-21
testing/firestore: another prefix test

jba:
Test the case where one path is a prefix of another, and the
values could in principle be combined.


schmidt-sebastian:
LGTM.

Is there some way we could version these? Having a version number on the binfile would make it much easier to see which client supports which test set.

jskeet:
I'm going to merge this myself so that I can easily start using the tests in the C# code.
(Sebastian's versioning idea is an interesting one, but let's handle that separately.)

jba:
I think the best version number is the commit hash. See https://code-review.googlesource.com/#/c/gocloud/+/20070 for my suggestion.

PR #241, opened 2018-01-05
testing/firestore: binary test file contains a single TestSuite proto

jba:
Change the format of the binary test file. It now contains a single
proto, TestSuite, that has a repeated field with all the tests. Now readers
of the binary file do not need to decode varints.

The file has a new name as well: test-suite.binproto.


schmidt-sebastian:
Thanks!

jskeet:
I don't understand Go so can't review the actual code. The plan sounds fine though. I'm away next week so won't be able to adjust the C# code until I get back, but that shouldn't cause any problems.

PR #242, opened 2018-01-10
testing/firestore: more tests for Set

jba:
Add tests, mostly for Set/merge.


schmidt-sebastian:
LGTM. Note that I already pulled these files into the Java and Node SDK.

jskeet:
Interestingly, this has introduced two test failures in C# - I need to dig deeper to figure out what's wrong. Just goes to show the value of the tests...

PR #249, opened 2018-03-27
testing/firestore: add tests for unary query filters

jba:
A where clause testing with null or NaN results in a unary query filter.
Add two tests for that.

Test interpreters need to interpret the JSON string "NaN" as a floating-point
NaN for the second test to work.


jba:
@schmidt-sebastian

schmidt-sebastian:
LGTM based on https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3091

jba:
> Do we already have tests that check that null and NaN are prohibited for other operators?

No. Will add.

PR #251, opened 2018-03-29
testing/firestore: an empty set with merge is now valid

jba:
Previously, we treated set({}, MergeAll) as an error. Now, it is valid.

See internal Google bug 73495873.


jba:
@schmidt-sebastian

jba:
Done.

jskeet:
I'm going to merge this so I can implement in C# :)

PR #253, opened 2018-04-26
testing/firestore: more watch tests

jba:


jba:
@schmidt-sebastian

jskeet:
I'm still not in a good position to review these - I think Sebastian is definitely more appropriate. Having said that, I'm *nearly* in a position where I can implement the tests in C#.

testing/firestore/testdata/listen-reset.textproto
```diff
@@ -0,0 +1,258 @@
...
+          key: "a"
+          value: <
+            integer_value: 3
+          >
+        >
+        create_time: <
+          seconds: 1
+        >
+        update_time: <
+          seconds: 3
```
schmidt-sebastian: Another case here is this very same document update with the original update_time of 2 seconds. At that point, we should mark this document as MODIFIED.

testing/firestore/testdata/listen-filter-nop.textproto
```diff
@@ -0,0 +1,247 @@
...
+        create_time: <
+          seconds: 1
+        >
+        update_time: <
+          seconds: 1
+        >
+      >
+      target_ids: 1
+    >
+  >
```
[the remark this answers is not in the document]
jba: I had that, but it's tricky. It causes the stream to close and reopen, and my test interpreter can't really handle that--it assumes the entire list of responses is returned from a single stream instance. 

If you find that it's easy to implement that in your node and Java implementations, then I'll add the test and fix my implementation somehow to make it work.

testing/firestore/testdata/listen-reset.textproto
jba: How could a document have a different contents but the same update time?

jba: How could a document have a different contents but the same update time?

schmidt-sebastian: Sorry I think I meant to say that we "should NOT mark the document as modified". I was thinking about a sequence of:

Doc A (version 1) -> Reset -> Doc A (version 1), Doc B

We should see two snapshots: One with Doc A added, and one with Doc B added. We need to add Doc B to make sure we get a new snapshot.

jba: OK, I expanded the test to do what you suggest (I think).

PR #259, opened 2018-08-01
testing/firestore: fix proto literals

jba:
Add field names to proto literals to deal with changes to the proto compiler.


blowmage:
Can you also update the `*.textproto` files? We noticed some of them were out of date compared to what the binary proto file contained.

jba:
This PR represents a consistent commit, where all the files (text and binary) were generated by the program. If you're not seeing a textproto here, that's because its contents didn't change. (Maybe it was the bin file that was out of date?) 

Which specific tests are different between text and binary?

blowmage:
I don't remember which test it was, but it was an update with either a server time or a delete, and the binproto said the mask had to be `['a', 'b']`, but the textproto file only had `['a']`. Perhaps I was looking at the wrong textproto.

¯\\\_(ツ)\_/¯

PR #267, opened 2018-10-09
chore: add issue templates

JustinBeckwith:
Mostly using this as a discussion prompt to talk about how we manage this across all repos.  This change will add a selector thingy that happens when you create a new bug on a repo.  Once we agree on the messaging, we need to figure out how to roll this out broadly.  

For an example, see:
https://github.com/googleapis/google-cloud-node/issues/new/choose

image

The key text added here is:

> **PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Bikeshedding on the words welcomed :)


JustinBeckwith:
cc @googleapis/yoshi

JustinBeckwith:
So ignore contributing - I just removed that.  The only other file with language or repo specific stuff is `bug_report.md`.  We can either a.) use synth templating to make this work like we're doing today with node.js, or b.) strip any language specific debugging/info/steps from the bug template.  I would really prefer to keep that in place, because it helps get things like node version, npm version, etc.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+---
+
+Thanks for stopping by to let us know something could be better!
+
+**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.
+
+Please run down the following list and make sure you've tried the usual "quick fixes":
+
+  - Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-common/issues
```
JustinBeckwith: This would be templated using synth to match the current repository url

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform
+  - Check our Troubleshooting guide: https://googlecloudplatform.github.io/google-cloud-node/#/docs/guides/troubleshooting
+  - Check our FAQ: https://googlecloudplatform.github.io/google-cloud-node/#/docs/guides/faq
+
+If you are still having issues, please be sure to include as much information as possible:
+
+#### Environment details
+
+  - OS:
```
JustinBeckwith: These would be templated as well, on a per-language basis

theacodes:
If that's the case, this PR needs to be against synthtool, not this repo.

JustinBeckwith:
I mean, that's the eventual plan, yeah.  I just wanted to start here to have the group discussion and agree on the text and specific issue types.  After we have agreement there, we can figure out how to assign out the work to templatize it.

PR #268, opened 2018-10-15
firestore: add conformance tests for empty cursor values

jadekler:
Per https://github.com/googleapis/google-cloud-dotnet/issues/2588, these are tests to ensure conformance around empty StartAt and EndBefore values.


jadekler:
cc @schmidt-sebastian

testing/firestore/testdata/query-cursor-endbefore-empty-array.textproto
```diff
@@ -0,0 +1,33 @@
...
+  >
+  query: <
+    from: <
+      collection_id: "C"
+    >
+    order_by: <
+      field: <
+        field_path: "a"
+      >
+      direction: ASCENDING
```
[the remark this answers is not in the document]
jadekler: Ahh ok, subtle. Could you PTAL the latest version? I _think_ I have it represented, but let me know if not.

blowmage:
The Ruby implementation raises `ArgumentError: must provide values`. Am I correct that these should be treated as a no-op? Therefor this code:

```ruby
require "google/cloud/firestore"

firestore = Google::Cloud::Firestore.new

# Get a collection reference
cities_col = firestore.col "cities"

# Create a query
query = cities_col.order(:population, :desc)
                  .order(:name)
                  .end_before()
```

Should be treated the same as:

```ruby
require "google/cloud/firestore"

firestore = Google::Cloud::Firestore.new

# Get a collection reference
cities_col = firestore.col "cities"

# Create a query
query = cities_col.order(:population, :desc)
                  .order(:name)
```

schmidt-sebastian:
> query = cities_col.order(:population, :desc) .order(:name) .end_before()

This should be rejected:

```
query = cities_col.order(:population, :desc)
                  .order(:name)
                  .end_before()
```

The following should be valid and uses an explicit empty map as a cursor. It is not the same as omitting it:

```
query = cities_col.order(:population, :desc)
                  .order(:name)
                  .end_before(EMPTY_MAP_IN_RUBY)
```

testing/firestore/testdata/query-cursor-endbefore-empty-array.textproto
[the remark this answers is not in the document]
schmidt-sebastian: You need to include the Map, even when it is empty. This is a test in Node:

```
  it.only('has startAt() method', () => {
    const ref1 = randomCol.doc('doc1');
    const ref2 = randomCol.doc('doc2');

    return Promise.all([ref1.set({foo: 'a'}), ref2.set({foo: {}})])
        .then(() => {
          return randomCol.orderBy('foo').endBefore({}).get();
        })
        .then(res => {
          console.log('Received ' + res.size + ' dcouemnts');
        });
  });
```

The output is 

```
Firestore (0.17.0) 2018-10-15T17:12:57.980Z ##### [Firestore]: Initialized Firestore
Firestore (0.17.0) 2018-10-15T17:12:57.982Z 4nGZ2 [WriteBatch.commit]: Sending 1 writes
Firestore (0.17.0) 2018-10-15T17:12:58.487Z ##### [Firestore]: Initialized Firestore GAPIC Client
Firestore (0.17.0) 2018-10-15T17:12:58.488Z DM5vH [WriteBatch.commit]: Sending 1 writes
Firestore (0.17.0) 2018-10-15T17:12:58.490Z ##### [Firestore._detectProjectId]: Detected project ID: node-review
Firestore (0.17.0) 2018-10-15T17:12:58.493Z 4nGZ2 [Firestore.request]: Sending request: {"database":"projects/node-review/databases/(default)","writes":[{"update":{"name":"projects/node-review/databases/(default)/documents/node_0.17.0_fEIygEfC6FbtPbCzdDiB/doc1","fields":{"foo":{"stringValue":"a"}}}}]}
Firestore (0.17.0) 2018-10-15T17:12:58.493Z DM5vH [Firestore.request]: Sending request: {"database":"projects/node-review/databases/(default)","writes":[{"update":{"name":"projects/node-review/databases/(default)/documents/node_0.17.0_fEIygEfC6FbtPbCzdDiB/doc2","fields":{"foo":{"mapValue":{}}}}}]}
Firestore (0.17.0) 2018-10-15T17:12:59.097Z DM5vH [Firestore.request]: Received response: {"writeResults":[{"transformResults":[],"updateTime":{"seconds":"1539623579","nanos":55183000}}],"commitTime":{"seconds":"1539623579","nanos":55183000}}
Firestore (0.17.0) 2018-10-15T17:12:59.100Z 4nGZ2 [Firestore.request]: Received response: {"writeResults":[{"transformResults":[],"updateTime":{"seconds":"1539623579","nanos":63564000}}],"commitTime":{"seconds":"1539623579","nanos":63564000}}
Firestore (0.17.0) 2018-10-15T17:12:59.103Z a6tV4 [Firestore.readStream]: Sending request: {"parent":"projects/node-review/databases/(default)","structuredQuery":{"from":[{"collectionId":"node_0.17.0_fEIygEfC6FbtPbCzdDiB"}],"orderBy":[{"field":{"fieldPath":"foo"},"direction":"ASCENDING"}],"endAt":{"values":[{"mapValue":{}}],"before":true}}}
Firestore (0.17.0) 2018-10-15T17:12:59.230Z a6tV4 [Firestore.readStream]: Received response: {"document":{"fields":{"foo":{"stringValue":"a","valueType":"stringValue"}},"name":"projects/node-review/databases/(default)/documents/node_0.17.0_fEIygEfC6FbtPbCzdDiB/doc1","createTime":{"seconds":"1539623579","nanos":63564000},"updateTime":{"seconds":"1539623579","nanos":63564000}},"transaction":{"type":"Buffer","data":[]},"readTime":{"seconds":"1539623579","nanos":179177000},"skippedResults":0}
Firestore (0.17.0) 2018-10-15T17:12:59.230Z a6tV4 [Firestore._initializeStream]: Releasing stream
Firestore (0.17.0) 2018-10-15T17:12:59.234Z a6tV4 [Firestore._initializeStream]: Received stream end
Received 1 documents

```

If I omit the empty map, I get 0 documents:

```
Firestore (0.17.0) 2018-10-15T17:15:25.492Z ##### [Firestore]: Initialized Firestore
Firestore (0.17.0) 2018-10-15T17:15:25.495Z UDh1h [WriteBatch.commit]: Sending 1 writes
Firestore (0.17.0) 2018-10-15T17:15:25.998Z ##### [Firestore]: Initialized Firestore GAPIC Client
Firestore (0.17.0) 2018-10-15T17:15:25.999Z NPiLh [WriteBatch.commit]: Sending 1 writes
Firestore (0.17.0) 2018-10-15T17:15:26.001Z ##### [Firestore._detectProjectId]: Detected project ID: node-review
Firestore (0.17.0) 2018-10-15T17:15:26.003Z UDh1h [Firestore.request]: Sending request: {"database":"projects/node-review/databases/(default)","writes":[{"update":{"name":"projects/node-review/databases/(default)/documents/node_0.17.0_FpyfMbWFyDXGR790rj68/doc1","fields":{"foo":{"stringValue":"a"}}}}]}
Firestore (0.17.0) 2018-10-15T17:15:26.004Z NPiLh [Firestore.request]: Sending request: {"database":"projects/node-review/databases/(default)","writes":[{"update":{"name":"projects/node-review/databases/(default)/documents/node_0.17.0_FpyfMbWFyDXGR790rj68/doc2","fields":{"foo":{"mapValue":{}}}}}]}
Firestore (0.17.0) 2018-10-15T17:15:26.610Z NPiLh [Firestore.request]: Received response: {"writeResults":[{"transformResults":[],"updateTime":{"seconds":"1539623726","nanos":566272000}}],"commitTime":{"seconds":"1539623726","nanos":566272000}}
Firestore (0.17.0) 2018-10-15T17:15:26.628Z UDh1h [Firestore.request]: Received response: {"writeResults":[{"transformResults":[],"updateTime":{"seconds":"1539623726","nanos":589683000}}],"commitTime":{"seconds":"1539623726","nanos":589683000}}
Firestore (0.17.0) 2018-10-15T17:15:26.632Z d8Tyt [Firestore.readStream]: Sending request: {"parent":"projects/node-review/databases/(default)","structuredQuery":{"from":[{"collectionId":"node_0.17.0_FpyfMbWFyDXGR790rj68"}],"orderBy":[{"field":{"fieldPath":"foo"},"direction":"ASCENDING"}],"endAt":{"before":true}}}
Firestore (0.17.0) 2018-10-15T17:15:26.752Z d8Tyt [Firestore.readStream]: Received response: {"document":null,"transaction":{"type":"Buffer","data":[]},"readTime":{"seconds":"1539623726","nanos":716261000},"skippedResults":0}
Firestore (0.17.0) 2018-10-15T17:15:26.752Z d8Tyt [Firestore._initializeStream]: Releasing stream
Firestore (0.17.0) 2018-10-15T17:15:26.756Z d8Tyt [Firestore._initializeStream]: Received stream end
Received 0 documents
```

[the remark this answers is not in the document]
jadekler: Done, I believe; PTAL

jdpedrie:
PHP will require a patch to conform. It's just one line of code though, nothing major. Thanks for the additional test! :)

testing/firestore/testdata/query-cursor-endbefore-empty-array.textproto
```diff
@@ -0,0 +1,40 @@
...
+  clauses: <
+    order_by: <
+      path: <
+        field: "a"
+      >
+      direction: "asc"
+    >
+  >
+  clauses: <
+    end_before: <
```
blowmage: I am very confused why the input is not providing the "explicit empty map". This is what I am expecting the setup to have in order to produce the results that are being asserted:

```
    end_before: <
      json_values: "[{}]"
    >
```

schmidt-sebastian: Oh, yes. This should be an empty map in the input. Providing nothing is invalid.

blowmage: It might be a good idea to have some additional empty input tests that raise errors as part of this change.

jadekler: Sorry for missing that. Quick clarification - you mention `json_values: "[{}]"`. Do you mean `json_values: "{}"`?

testing/firestore/testdata/query-cursor-endbefore-empty-array.textproto
```diff
@@ -0,0 +1,41 @@
+# DO NOT MODIFY. This file was generated by
+# github.com/GoogleCloudPlatform/google-cloud-common/testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go.
+
+# Cursor methods are allowed to use empty values with EndBefore. It should result
+# in no addition to the query.
+
+description: "query: EndBefore with empty values"
```
blowmage: Shouldn't this be:

```
description: "query: EndBefore with explicit empty map"
```

testing/firestore/testdata/query-cursor-startat-empty-array.textproto
```diff
@@ -0,0 +1,41 @@
+# DO NOT MODIFY. This file was generated by
+# github.com/GoogleCloudPlatform/google-cloud-common/testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go.
+
+# Cursor methods are allowed to use empty values with StartAt. It should result in
+# no addition to the query.
+
+description: "query: StartAt with empty values"
```
blowmage: Shouldn't this be:

```
description: "query: StartAt with explicit empty map"
```

blowmage:
These tests now specify the behavior of passing an empty map object, and not empty values. From the conversation, I would assume a test case that looks like this:

```
description: "query: StartAt does not accept empty values"
query: <
  coll_path: "projects/projectID/databases/(default)/documents/C"
  clauses: <
    order_by: <
      path: <
        field: "a"
      >
      direction: "asc"
    >
  >
  clauses: <
    start_at <
    >
  >
  is_error: true
>
```

testing/firestore/testdata/query-cursor-endbefore-empty-array.textproto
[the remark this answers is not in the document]
jskeet: The current version, with `map_value: <>` contains a value, which isn't the same as not having *any* values in the `end_at`. This doesn't trip the bug in my library at the moment.

It may well be an entirely valid test, but we need *another* test where we don't specify any values at all.

blowmage:
Are there more changes needed for this PR?

jskeet:
Yes - as it is, this doesn't trip my previous bug. As per my comment: 

> The current version, with `map_value: <>` contains a value, which isn't the same as not having any values in the `end_at`. This doesn't trip the bug in my library at the moment.

> It may well be an entirely valid test, but we need another test where we don't specify any values at all.

jadekler:
Hey y'all, sorry this fell off my radar there, and thanks for pinging it. I admit that I'm fairly confused about whether the test should be,

```
clauses: start_at: json_values: ""
(results in)
error
```

Or,

```
clauses: start_at:
(results in)
error
```

@schmidt-sebastian suggested the former earlier this week in a chat, and [@blowmage the latter](https://github.com/googleapis/google-cloud-common/pull/268#issuecomment-430099474). Thoughts from y'all?

schmidt-sebastian:
> @schmidt-sebastian suggested the former earlier this week in a chat, and [@blowmage the latter](https://github.com/googleapis/google-cloud-common/pull/268#issuecomment-430099474). Thoughts from y'all?

```
.startAt(undefined) -> not allowed
.startAt() -> not allowed
.startAt({}) -> allowed
```

I believe that the test tooling treats as "" (which is not valid JSON) as "{}". "{}" => {} and as such it is allowed.

FWIW, I believe that at least the Node client handles this correctly, so you can play with it to tune your test cases.

blowmage:
Thanks for updating the tests. Ruby passes using the updated binary file. LGTM. :+1:

tseaver:
I just verified that the Python unit tests pass with the four new `.textproto` files:

```bash
$ nox -re unit-3.6 -- -s -k test_cross_language -vv | grep "query-cursor-.*-empty"
...
 - tests/unit/testdata/query-cursor-endbefore-empty-map.textproto
 - tests/unit/testdata/query-cursor-endbefore-empty.textproto
 - tests/unit/testdata/query-cursor-startat-empty-map.textproto
 - tests/unit/testdata/query-cursor-startat-empty.textproto
nox > Session unit-3.6 was successful.
```.

**Comment by jskeet** [comment]
I'm going to be away for a few days, and it's harder for me to test before it's merged. If everyone else is happy, I suggest you now merge, and then I check afterwards that it tests what I expect :)

**Comment by jskeet** [comment]
I've only just got round to merging this, and it fails in C# for the "EndBefore with empty values" and "StartAt with empty values" tests. Those are marked as being errors, but I believe they should be allowed. This was precisely what I originally requested the original tests for, when fixing https://github.com/googleapis/google-cloud-dotnet/issues/2588

The issue was specifically about passing in an empty set of values, *not* "a set of values with one element, which is an empty map". So that's effectively your `.startAt()` case. 

@schmidt-sebastian Could you double-check the intention here? We need to make sure we're really consistent on this. In an email thread it sounded like you *did* want to allow this.

**Comment by schmidt-sebastian** [comment]
I just updated the tests in Node and Java. I actually had to add validation to reject empty calls to StartAt(). Both SDKs allowed them (well, Node didn't ... but it failed with an assertion in the GRPC layer).

The example in #2588 is
```
var anomaly = await firestoreDB.Value.Collection("col").OrderByDescending("field").StartAt(new object[0])
```

This is different from 
```
var anomaly = await firestoreDB.Value.Collection("col").OrderByDescending("field").StartAt()
```

#2588 should map to a cursor value of  `{ mapValue: {} }`. I believe that the conformance tests in this PR are correct here.

jskeet:
@schmidt-sebastian: The conformance tests here prevent `StartAt(new object[0])` from working. That's saying "I have no starting values" not "the first StartAt field is an empty map".

Basically, we had a restriction in C#, a user asked us to remove the restriction, and now we've got conformance tests that require the restriction, as far as I can see :(

schmidt-sebastian:
Hm, then I missed the nuance in the C# syntax when I commented on #2588. Supporting an empty `startAt` is not the end of the world - but it doesn't add any value over not specifying a `startAt` at all. 

I am not sure on what the best path forward is - is there a way you can exclude these tests? I don't want to break your users. Alternatively, we could allow this everywhere, but calling this API with no values doesn't actually specify a cursor.

Sorry for missing this nuance.

jskeet:
@schmidt-sebastian: Glad we're on the same page now :) It can be tricky describing things with so many languages involved.

While I *could* exclude the tests, I think it would be useful to allow this everywhere, to be honest - unless it's hard to explain the actual behaviour. I'd assume it would be equivalent to not specifying a starting point; the OP in 2588 sounds like they expect something with possibly more nuance:

> Actually it should let me got the whole documents which has value of that field not an array nor object?

I can imagine cases where you may have a variable number of elements for the starting point depending on other factors, and allowing that variable number to be 0 without having to write special code to omit the `StartAt` call feels useful to me. But I don't know nearly as much about the use cases as you do.

Would it be helpful to get the OP of 2588 to discuss it further here?

jskeet:
@schmidt-sebastian: Ping for further discussion :) (Happy to do that internally too if that works better for you.)

schmidt-sebastian:
> @schmidt-sebastian: Ping for further discussion :) (Happy to do that internally too if that works better for you.)

I don't really have a good solution. Meeting scheduled.

PR #271, opened 2019-01-11
DO NOT MERGE firestore: switch from apiv1beta1 to apiv1

jadekler:


jadekler:
cc @blowmage @tseaver @schmidt-sebastian

jskeet:
I've tested this against a version of the C# code that uses V1, and it works fine.

blowmage:
Ruby just added V1. Is this ready to be used?

jadekler:
@blowmage I think so, but it does not include any of the conformance test modifications we might need. I'm working on those now. Happy to merge this to unblock y'all though if you'd like!

blowmage:
The Ruby conformance tests currently pass using this new binary file, so we don't need to make any other changes. Are there additional changes you are working on that will affect the binary file? New or changed tests?

jadekler:
Ah, SG. Let's move this forward then whilst I investigate conformance tests.

PR #275, opened 2019-02-06
add a nested ServerTimestamp testcase

jadekler:
From: https://github.com/googleapis/google-cloud-python/issues/7215#issuecomment-461175998


jadekler:
@tseaver Does this look right?

cc @mikelehen

testing/firestore/testdata/update-st-nested-dot.textproto
```diff
@@ -0,0 +1,43 @@
+# DO NOT MODIFY. This file was generated by
+# github.com/GoogleCloudPlatform/google-cloud-common/testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go.
+
+# TODO(tres)
+
+description: "update: update: nested ServerTimestamp field via dotted names"
```
mikelehen: Is "update: update:" expected?

testing/firestore/testdata/update-paths-st-nested-dot.textproto
```diff
@@ -0,0 +1,47 @@
...
+              >
+            >
+          >
+        >
+      >
+      current_document: <
+        exists: true
+      >
+    >
+    writes: <
```
mikelehen: Why is "writes:" duplicated?  Is this representing a write stream or something? Or is this just how arrays end up represented in textproto?

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,17 @@ update operation should be produced.`,
 			maskForUpdate: nil,
 			transform:     transforms(st("a")),
 		},
+		{
+			suffix: "st-nested-dot",
+			desc:   "update: nested ServerTimestamp field via dotted names",
+			comment: `TODO(tres)`,
```
mikelehen: tres?

testing/firestore/testdata/update-st-nested-dot.textproto
jadekler: Fixed - thanks!

testing/firestore/testdata/update-paths-st-nested-dot.textproto
jadekler: I _think_ it's just how arrays are represented in textproto.

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
jadekler: @tseaver could you provide a comment about this test?

tseaver: Per https://github.com/googleapis/google-cloud-python/issues/7215, I don't know what the expected behavior is for the case where a dotted name has a server timestamp transform.  The actual behavior of the Python library is to send along field masks for (in the case of the data below) `a.b` and `a`, which causes any existing fields inside `a` to get wiped out, rather than merging.  I'm not sure where the requirement that we pass the "parent" of the nested server timestamp key comes from, but I don't believe that is a desirable behavior.

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,17 @@ update operation should be produced.`,
...
 		},
+		{
+			suffix: "st-nested-dot",
+			desc:   "nested ServerTimestamp field via dotted names",
+			comment: `TODO(tres)`,
+			inData:        `{"a.b": 1, "a.c": "ServerTimestamp"}`,
+			paths:         [][]string{{"a.b", "a.c"}},
+			values:        []string{`[1, "ServerTimestamp"]`},
+			outData:       mp("a", mp("b", 1)),
+			maskForUpdate: nil,
```
tseaver: I don't really know the semantics of the generator here:  do we really mean to be passing no field mask at all?  That certainly seems to be so in the textproto below.  The [docs for the `update_mask` field](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1beta1#google.firestore.v1beta1.Write) state:

> If the mask is not set for an update and the document exists, any existing data will be overwritten.

PR #276, opened 2019-02-07
adds a test for updating nested values

jadekler:
From: https://github.com/googleapis/google-cloud-python/issues/7215

It's currently possible to send an update like:

```
- (json Update version): {"a.b": 7, "a.c": "ServerTimestamp"}
- (fieldpath UpdatePath version):
    path: "a.b"
    value: 7
    path: "a.c"
    value: "ServerTimestamp"
```

This is saying: update key "b" within key "a" with value 7, and
vice-versa for key "c" within key "a" with value "ServerTimestamp".

This particular conformance test ensures that the update that gets sent
- especially in the json case - is tightly scoped to a.b, and that a
field mask is not added for "a". Currently, a bug in python does the
latter - see
googleapis/google-cloud-python#7215.
That is, it sends an update like:

```
path: a.b
value: 7
path: a.c
value: ServerTimestamp
```

Instead of:

```
path: a
value: {"b": 7}
```

Or:

```
path: a
value: {"b": "ServerTimestamp"}
```

They look the same, but the former _only_ updates a.b to 7, whereas the
latter replaces the entirety of a with {"b": 7}.


jadekler:
@tseaver Could you check out the description + generated .textproto and see if this fits better? I _think_ I understand your problem better now. I removed all ServerTimestamp parts, since it seems unrelated to the underlying problem. Let me know if I've misunderstood the issue in https://github.com/googleapis/google-cloud-python/issues/7215 though.

tseaver:
@jadekler Can you please link to where the semantics are specified for `update`?  I have no clue how users of our client libraries are supposed to figure them out, frankly, since they are pretty much unrelated to anything in the back-end API docs for `CommitRequest`, `Write`, etc.  The only relevant docs I can see are [here](https://cloud.google.com/firestore/docs/manage-data/add-data#update_fields_in_nested_objects), and they don't specify the behavior.

jadekler:
For posterity: continuing conv offline.

jadekler:
cc @mikelehen

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,28 @@ update operation should be produced.`,
...
+			inData:        `{"a": {"b": 7}}`,
+
+			// paths => field_paths (paths) => []firestore.Update.Path
+			paths:         [][]string{{"a.b"}},
+			// inData => json_values => []firestore.Update.Val
+			values:        []string{`7`},
+			// outData => request.writes => request expectation
+			outData:       mp("a.b", 7),
+			// maskForUpdate => request.update_mask => request expectation
+			maskForUpdate: []string{"`a.b`"},
```
mikelehen: Looking at this test, I'm either confused or this is not the behavior I expect.  Taking node.js as an example, `doc.update({a: {b: 7} })` will completely overwrite the field "a" (i.e. using an update_mask of ['a']) which seems to be different than what this test is expecting, right?

FYI- To update a.b, you'd do `doc.update({'a.b': 7})` instead.  Or we also allow `doc.update('a.b', 7)` or `doc.update(firestore.FieldPath('a', 'b'), 7)`.

jadekler: That's makes sense to me. I can update this json to {a.b: 7} instead of {a: {b: 7}}.

The json portion of the test is used for Update tests, whereas the paths+values portion is used for UpdatePath. In Go, we only use UpdatePath, so I was only able to verify the latter. 

@mikelehen Before I make this change, does the rest of this test look correct?

@tseaver before I make this change, does that match your expectation?

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,28 @@ update operation should be produced.`,
...
 			transform:     transforms(st("a")),
 		},
+		{
+			suffix: "nested-single-value",
+			desc: 	"Updating a nested value results in queries that are tightly" +
+"scoped to that specific field.",
+			comment:   "Changing a.b sends an update that's scoped specifically to" +
+"a.b, instead of sending an update that changes the entirety of a. For example," +
+"its field_key should be a.b: 7, not a: b: 7 (which would entirely replace all of" +
+"a and blow away anything other than a.b).",
```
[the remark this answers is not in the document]
jadekler: Done

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,28 @@ update operation should be produced.`,
 			maskForUpdate: nil,
 			transform:     transforms(st("a")),
 		},
+		{
+			suffix: "nested-single-value",
+			desc: 	"Updating a nested value results in queries that are tightly" +
```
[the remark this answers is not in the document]
jadekler: Done

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
wilhuff: Discussed via chat: updating to {"a.b": 7} is right.

jadekler: Done. @tseaver, will leave the verification as to whether this test covers your issue to you.

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,28 @@ update operation should be produced.`,
...
+		{
+			suffix: "nested-single-value",
+			desc: 	`Updating a nested value results in update masks that are tightly 
+scoped to that specific field.`,
+			comment:   `Changing a.b sends an update that's scoped specifically to
+a.b, instead of sending an update that changes the entirety of a. For example,
+"its field_key should be a.b: 7, not a: b: 7 (which would entirely replace all of
+"a and blow away anything other than a.b).`,
+			// inData => json_data (non-paths) => ???
+			inData:        `{"a.b": 7}}`,
```
blowmage: This is invalid JSON.

blowmage:
This PR looks like a change in behavior to me. The update method currently takes the values given as they are. So `"a.b": 7` would set the field `a.b` to `7`. The way to specify a nested structure is to pass a FieldPath. In Ruby that would be `doc.update(firestore.field_path(:a, :b) => 7)`.

Update treats the data passed to it differently than Set. Update does not treat the fields names like field paths as Set does. If you pass `"a.b": 7` to Set it will consider the data `{ a: { b: 7 } }`. But if you pass `"a.b": 7` it will consider the data `{ "a.b": 7 }`.

Update also only considers the top-level fields. So if you pass `{ a: { b: 7 } }` to Update it will replace all of `a` with ` { b: 7 }`. Whereas if you pass `{ a: { b: 7 } }` to Set it will consider it the same as passing a field path of `a.b` with a value of `7`. AFAIK these two different behaviors are intentional and described in the design document.

tseaver:
@blowmage Thanks for articulating that conflict.  I'd really like for the semantics set, set-with-merge, and update to have, explicit, thorough, and *public* documentation:  how can we expect users to deduce them, when we can't even get them right?

blowmage:
I agree that the behavior is confusing. In fact, I think I was confused and conflated what Set and Update do. Here is the `#update` description in the ruby documentation:

> The top-level keys in the data hash are considered field paths, and can either be a FieldPath object, or a string representing the nested fields. In other words the string represents individual fields joined by ".". Fields containing `~`, `*`, `/`, `[`, `]`, and `.` cannot be in a dotted string, and should provided using a `FieldPath` object instead.

So I was wrong in my previous comment. There are two different behaviors, but Update treats `"a.b"` as a field path, and Set treats `"a.b"` as a field name. Basically, Update splits on dots, but Set does not.

testing/firestore/testdata/update-nested-single-value.textproto
```diff
@@ -0,0 +1,33 @@
...
+
+# Changing a.b sends an update that's scoped specifically to a.b, instead of
+# sending an update that changes the entirety of a. For example, "its field_key
+# should be a.b: 7, not a: b: 7 (which would entirely replace all of "a and blow
+# away anything other than a.b).
+
+description: "update: Updating a nested value results in update masks that are tightly \nscoped to that specific field."
+update: <
+  doc_ref_path: "projects/projectID/databases/(default)/documents/C/d"
+  json_data: "{\"a.b\": 7}}"
```
tseaver: Invalid JSON, extra `}`.

testing/firestore/testdata/update-nested-single-value.textproto
```diff
@@ -0,0 +1,33 @@
...
+        fields: <
+          key: "a.b"
+          value: <
+            integer_value: 7
+          >
+        >
+      >
+      update_mask: <
+        field_paths: "`a.b`"
+      >
```
tseaver: This output does not match the semantics implied by [`update-split-top-level.textproto`](https://github.com/googleapis/google-cloud-common/blob/master/testing/firestore/testdata/update-split-top-level.textproto) or [`update-split.textproto`](https://github.com/googleapis/google-cloud-common/blob/master/testing/firestore/testdata/update-split.textproto):

- Top-level keys are split, so that the `fields` in the `update` message should look like:

```protobuf
     fields {
       key: "a"
       value {
         map_value {
           fields {
             key: "b"
             value {
               integer_value: 7
             }
           }
         }
       }
     }
```

The `field_paths` in the update mask should not be quoted:

```protobuf
   update_mask {
     field_paths: "a.b"
   }
```

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
jadekler: What's invalid about it? It seems perfectly legal to me, but I might be missing something.

Also, I plan to rename these from "json" to "json-like", which I think is the more accurate description of curly brackets in node. Let me know if there's a more canonical description.

wilhuff: The value in backticks has unbalanced curly brackets.

blowmage: > Also, I plan to rename these from "json" to "json-like"

I don't care as much about what it is named, I care that it is parseable by a JSON library.

```ruby
JSON.parse '{"a.b": 7}}'  #=> JSON::ParserError (765: unexpected token at '}')
```

jadekler: Oh I see, there's too many curlys at the end. Fixed!

testing/firestore/testdata/update-nested-single-value.textproto
jadekler: Thank you! Fixed!

jadekler:
Just as a heads up, I finally have time to page this back in. Working on investigating y'alls comments. Sorry for the slowness getting back to this.

testing/firestore/testdata/update-nested-single-value.textproto
jadekler: cc @mikelehen @wilhuff 

TIL "only top-level keys are split". Could one of you speak to that? If that is correct, then @tseaver is correct that the consistent thing to do would be to treat `update({"a.b": 7})` as `update({"a": {"b": 7}})`, since:

> [The Update method splits [...] top-level keys at dots.](https://github.com/googleapis/google-cloud-common/blob/49080d7486e9c66b99d07a4a39406b019454c783/testing/firestore/testdata/update-split-top-level.textproto#L4)

--------------------------

For posterity: `update({"a": {"b": 7}})` replaces "a" with `{"b": 7}`, whereas `update(FieldPath: "a.b", Value: 7)` replaces "a.b" with 7.

wilhuff: `update({"a.b": 7})` is not the same as `update({"a": {"b": 7}})`

The former replaces only b within a. The latter replaces the entirety of a.

jadekler:
@tseaver @blowmage I've gone through this once more with @mikelehen, and we've corrected a few things. Minus the random documentation here and there, could you two take a peek through this and see if it matches your expectation?

jadekler:
@tseaver @blowmage I searched through the ruby and python docs. AFAICT the dotted-path behavior is fairly explicitly called out.

--------------

## Python

https://googleapis.github.io/google-cloud-python/latest/firestore/document.html

screen shot 2019-03-04 at 2 46 23 pm

--------------

## Ruby

http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-firestore/latest/Google/Cloud/Firestore/DocumentReference.html#update-instance_method

screen shot 2019-03-04 at 2 46 41 pm

--------------

Could y'all narrow in the parts of update / set that you'd like to see better documentation around? FWIW it basically boils down to the fact that Update({"a.b": 7}) results in an update to field "b" in field "a", whereas Set({"a.b": 7}) results in the overwrite of field "a.b".

blowmage:
It looks like the `test-suite.binproto` file in this PR has been updated, and it passes in Ruby as-is. :+1: for me.

tseaver:
Python passes with the added `update-nested-single-value.textproto` test with no changes.  However, I don't believe that this new test actually addresses the issue in https://github.com/googleapis/google-cloud-python/issues/7215:  there, the addition of a transform as a dotted key causes the entire mapping to be replaced, because of an (inferred) rule in Python that nested transforms in an update require includng the parent in the `update_mask`.  E.g., for the following update data:

```
  json_data: "{\"a.b\": 7, \"a.c\":  \"ServerTimestamp\"}"
```

the current Python implementation yields both and `update` and a `transform` for the document, and *includes `a` in the `update_mask`*:

```
    writes: <
      update: <
        name: "projects/projectID/databases/(default)/documents/C/d"
        fields: <
          key: "a"
          value: <
            map_value: <
              fields: <
                key: "b"
                value: <
                  integer_value: 7
                >
              >
            >
          >
        >
      >
      update_mask: <
        field_paths: "a.b"
        field_paths: "a"
      >
      current_document: <
        exists: true
      >
    >
    writes: <
      transform: <
        document: "projects/projectID/databases/(default)/documents/C/d"
        field_transforms: <
          field_path: "a.c"
          set_to_server_value: REQUEST_TIME
        >
      >
      current_document: <
        exists: true
      >
    >
```

I *believe* that this behavior is incorrect, and that it should instead omit `a`:

```
    writes: <
      update: <
        name: "projects/projectID/databases/(default)/documents/C/d"
        fields: <
          key: "a"
          value: <
            map_value: <
              fields: <
                key: "b"
                value: <
                  integer_value: 7
                >
              >
            >
          >
        >
      >
      update_mask: <
        field_paths: "a.b"
      >
      current_document: <
        exists: true
      >
    >
    writes: <
      transform: <
        document: "projects/projectID/databases/(default)/documents/C/d"
        field_transforms: <
          field_path: "a.c"
          set_to_server_value: REQUEST_TIME
        >
      >
      current_document: <
        exists: true
      >
    >
```

I don't understand the requirement to include the parent of a dotted-transform key in the update mask:  it may be that the Python implementation confuses the dotted version above with a version using an actual mapping for a value:

```
  json_data: "{\"a\": {\"b\": 7, \"c\":  \"ServerTimestamp\"}}"
```

schmidt-sebastian:
@tseaver Your `believe` is correct. The current output `update_mask:['a.b', 'a']` is treating by the backend as 'update_mask:a' and replaces all values under `a`. For your example, you must explicitly set the field mask to `a.b`. The transform can then be applied to `a.c` only.

tseaver:
@schmidt-sebastian Thanks for the clarification.

@jadekler Can we amend this PR to add a test which exercises the mixed update / transform case where both are dotted keys in the same top-level mapping (i.e, one which Python's current implementation would fail?)

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -239,6 +239,51 @@ update operation should be produced.`,
...
+			// unnecessary?
+			transform:     nil,
+		},
+		{
+			suffix: "nested-single-value-with-transform",
+			desc: 	`TODO.`,
+			comment:   `TODO.`,
+
+			// inData => json_data (non-paths) => ???
+			inData:        `{"a.b": "ServerTimestamp"}`,
```
tseaver: I'd prefer that this testcase mixed both a server timestamp *and* a normal subkey, e.g.:
```suggestion
			inData:        `{"a.b": 7, "a.c": "ServerTimestamp"}`,
```

jadekler: @tseaver Thanks for your comments. This is WIP. I'll ping when it's ready for re-review :) Also, I'm trying to reach you on hangouts if you're available?

jadekler:
Amended so that we include a ServerTimestamp in the test. PTAL.

schmidt-sebastian:
LGTM. Thanks for the update.

blowmage:
Ruby passes using the updated binproto file?

tseaver:
I can confirm that `update-nested-transform-and-nested-value.textproto` does indeed surface the bug from https://github.com/googleapis/google-cloud-python/issues/7215.

PR #279, opened 2019-03-06
Restore 'v1beta1' version of firestore test proto.

tseaver:
Needed to allow continued testing of the 'v1beta1' code in Python.

Renames 'test.proto' -> 'test_v1.proto', but leaves a 'test.proto' symlink
to allow scripts which do not need both to continue to work.

Adds a version qualifier to the protobuf package (so that descriptors don't
clash in a process which uses both files).


jadekler:
Could you describe why v1beta1 needs additional testing? Ideally we'd be on track to remove it entirely in favour of switching to v1.

tseaver:
@jadekler We don't plan to remove `v1beta1` from the Python package until after having released with both versions several months (quarters?), to allow users who need to fall back to the `v1beta1` version to do so without pinning the library.  /cc @crwilcox

In order to run the conformance tests against both versions, I need to be able to generate versioned `_pb2.py` files (with different descriptors).

tseaver:
For reference, see the Python V1 PR: https://github.com/googleapis/google-cloud-python/pull/7494.

jadekler:
Thanks for the notes and the link. Couple notes / questions about that:

- Currently the rough plan across languages is to only include both versions in a single release, and then remove it after-the-fact. More discussion to happen after languages have baked.
- After https://github.com/googleapis/google-cloud-python/pull/7494, I would assume that the only way that users in latest version can interact with v1beta1 is by using the gapic manually. Is that correct? That is, the manual will use v1 under the hood, not v1beta1, and inputs to the manual are v1 objects (not sure if that distinction is valuable in dynamically-typed languages).
- Given the last point, my expectation would be that https://github.com/googleapis/google-cloud-python/pull/7494 includes the new conformance tests (that use v1, not v1beta). (I see `firestore/tests/unit/v1/testdata/test-suite.binproto` in the PR, which miiiight be that?)

Since the conformance tests test the manual, and that in that PR the manual changes to v1, I'm having a bit of a hard time understanding how v1beta1 conformance tests would be helpful. Are y'all forking the manual instead of editing it in place, perhaps?

tseaver:
@jadekler:

- Re: removing `v1beta1` after a single release:  that would be fine with me, although that would be a "big" release, and we could in theory need to make a bugfix release or more before that one.
- https://github.com/googleapis/google-cloud-python/pull/7494 forks the manual code, not just the GAPIC, which is why we need the conformance tests for both versions:  the point is to let people fall back to `v1beta` just by changing their imports.
- The `binproto` file is just a copy of the `v1beta` version, which is itself a fossil:  Python doesn't actually use the binary file at all.  I don't believe the actual `.textproto` files have changed.

jadekler:
Gotcha. Thanks for the context. I remain OK with removing the v1beta1 conformance tests, since it should be frozen => removed shortly. However, if you feel strongly about this, I'm happy to move forward.

If you'd like to move forward, please fix the `make` command: currently it fails with this PR: https://gist.github.com/jadekler/542e4af3e2532b621de2532e1a8a9dab

jadekler:
Friendly ping - is this PR still on your radar?

PR #280, opened 2019-03-11
greatly expand on writetest documentation

jadekler:
It was pretty hard to understand, so here's filling out a bunch more
details that hopefully will make life easier for the next developer.


jadekler:
cc @schmidt-sebastian

jadekler:
No prob! Sorry, this wasn't meant as a criticism of previous work. I should have worded my message better. It may be that I work better when things are very verbosely spelled out heh.

PR #282, opened 2019-03-13
adds storage conformance tests

jadekler:
This is a proof of concept.


blowmage:
I would prefer to see a single JSON file output than having to load multiple files.

jskeet:
I concur on the single file aspect - not so much for avoiding loading multiple files, but for clarity. I'd generally approach data-oriented tests in the same way as regular unit tests - if two unit tests would live in the same source file, then ideally the corresponding data-oriented tests should live in the same source file. (Where practical, of course.)

Where there are multiple "kinds" of test (e.g. V4 signing tests and some entirely different kind of test) I don't have a strong opinion about whether each kind of test should have its own proto file - I'd probably *suggest* a separate message for each kind of test, but within the same proto.

I haven't yet looked at the files themselves - will do that now.

testing/storage/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/golang/protobuf/jsonpb"
+	pb "github.com/googleapis/google-cloud-common/testing/storage/genproto"
+)
+
+var testcases = []pb.SignedV4Test{
```
jskeet: Is this only a temporary measure? We shouldn't require users who want to add new tests to have Go available. Ideally, I'd like to see no code at all within this repo - it should *just* be the data and .proto files.

testing/storage/testdata/customer-supplied-encryption-key.json
```diff
@@ -0,0 +1,27 @@
+{
+	"fileName": "customer-supplied-encryption-key",
+	"description": "Customer-supplied encryption key",
+	"bucket": "test-bucket",
+	"object": "test-object",
+	"method": "GET",
+	"expiration": "10",
+	"timestamp": "20190201T090000Z",
+	"expectedUrl": "https://storage.googleapis.com/test-bucket/test-object?X-Goog-Algorithm=GOOG4-RSA-SHA256\u0026X-Goog-Credential=test-iam-credentials%40dummy-project-id.iam.gserviceaccount.com%2F20190201%2Fauto%2Fstorage%2Fgoog4_request\u0026X-Goog-Date=20190201T090000Z\u0026X-Goog-Expires=10\u0026X-Goog-SignedHeaders=host\u0026X-Goog-Signature=95e6a13d43a1d1962e667f17397f2b80ac9bdd1669210d5e08e0135df9dff4e56113485dbe429ca2266487b9d1796ebdee2d7cf682a6ef3bb9fbb4c351686fba90d7b621cf1c4eb1fdf126460dd25fa0837dfdde0a9fd98662ce60844c458448fb2b352c203d9969cb74efa4bdb742287744a4f2308afa4af0e0773f55e32e92973619249214b97283b2daa14195244444e33f938138d1e5f561088ce8011f4986dda33a556412594db7c12fc40e1ff3f1bedeb7a42f5bcda0b9567f17f65855f65071fabb88ea12371877f3f77f10e1466fff6ff6973b74a933322ff0949ce357e20abe96c3dd5cfab42c9c83e740a4d32b9e11e146f0eb3404d2e975896f74",
+	"headers": {
```
jskeet: One advantage of the JSON files being hand-written instead of generated by code (whether Go or anything else) would be that we could format them considerably more concisely :)

testing/storage/testdata/customer-supplied-encryption-key.json
```diff
@@ -0,0 +1,27 @@
+{
+	"fileName": "customer-supplied-encryption-key",
```
jskeet: I'd *hope* this could go away if we just go for JSON being the single source of truth, rather than Go. We may or may not need a "test ID" which is a valid identifier - I don't think we would for .NET, but it's possible that some languages would need parameterized tests to have distinct IDs.

testing/storage/README.md
```diff
@@ -0,0 +1,3 @@
+# Generating tests
+
+`make`
```
jskeet: This is unfriendly for Windows users who may not have Make, or protoc, or Go installed.

testing/storage/proto/testcase.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package tests;
+
+message HeaderList {
```
jskeet: To me this doesn't look like a list of headers, so much as values. I'd call it ValueList and have repeated string values. I'd still keep the *map* named headers, as it's the combination of "key + values" that determines each header.

testing/storage/README.md
jadekler: Thanks for the feedback. The makefile can certainly be rewritten to be a bash script (I think powershell lets you run shell scripts?). I chose to use makefiles because it was easy to copy paste heh. :)

I don't think protoc or Go have any windows specific issues? These are just dependencies of the current approach. Similarly, users using releasetool have to install pip and python https://github.com/googleapis/releasetool#installation.

testing/storage/testdata/customer-supplied-encryption-key.json
jadekler: +1, that's true

testing/storage/main.go
jadekler: I don't see a way to have no code with this approach. We could hand-write proto files and then hand-write json files, but then we don't get the specification guarantees.

That leaves us with having to have _some_ language in this approach. I'm pretty open to any language, but Go happens to be quite easy to install and run (download, run installer, run `go run main.go`) (the $GOPATH shenanigans have been eradicated - hooray!). But, yeah, not super opinionated on language.

testing/storage/README.md
jskeet: But why include any Go here at all? Why not just have the .proto files and the json files? Note that this means that in order to add/modify a test, I need to either know Go or be happy to fake it cargo-cult style. (It's not that I have anything against Go specifically - I'd expect the same push-back if I included code written in C# here that had to be modified in order to add a platform-neutral conformance test.)

(I'd suggest that the set of users using releasetool is narrower than the potential set of users who might want to add conformance tests - if not for Storage, then potentially for other APIs, if we're trying to work out a style that will be broadly applicable.)

If the concern is that the JSON might not be valid for the protos, we could add a CI test for that, which *could* depend on Go.

testing/storage/testdata/customer-supplied-encryption-key.json
jskeet: No, I'm not suggesting a non-proto approach. I'm suggesting that a proto approach doesn't need to involve writing code in order to represent new tests.

I'm suggesting that in this repo we have:

- JSON files
- Proto files
- CI tasks that validate that the JSON files are valid for the proto files

That keeps the benefits of protos in terms of being validated (and with statically-typed access in the test code itself, for suitable languages), but without *code* being used to add *data*.

jskeet: If anyone is uncomfortable with the JSON format for a proto, they can easily write code in *their* preferred language, instantiate the test object, then convert it to JSON and copy/paste that. But they don't *have* to do that, and there's no redundancy.

testing/storage/proto/testcase.proto
jadekler: Done, thanks for the good suggestion

jadekler:
> I would prefer to see a single JSON file output than having to load multiple files.

> I concur on the single file aspect - not so much for avoiding loading multiple files, but for clarity. 

Done

testing/storage/testdata/customer-supplied-encryption-key.json
jadekler: > CI tasks that validate that the JSON files are valid for the proto files

TIL that's something you can do. Thanks for following up. Could you point me at some reading for verification? That sounds very cool.

And, +1, that seems like a very good idea.

testing/storage/main.go
jadekler: Conv continued in https://github.com/googleapis/google-cloud-common/pull/282#discussion_r265441007 in which it turns out that I had no idea you could verify json files against a .proto specification, which seems like a very good way to go. I'll resolve this in favour of that conversation ( feel free to unresolve if there's an aspect of this conversation that's distinct from the other ).

testing/storage/testdata/customer-supplied-encryption-key.json
jskeet: I don't have any particular document - but it's basically "try to parse the JSON files" :)
In some languages this could potentially be done dynamically - steps of:

- Run protoc on test.proto to generate a descriptor set
- Load the descriptor set dynamically
- Parse with reflection

Alternatively, you could easily have some Go/Java/Python/whatever that ran protoc to generate the right code, and then load the file in exactly the same way that the tests would (e.g. in C# it would be something `SigningTest.Parser.ParseJson(json)`).
We'd probably want some sort of manifest to say which proto message was contained within each JSON file, but that should be easy enough to do (and would rarely change).

The one tricky part thinking off the top of my head is working out how to handle the "repeated" aspect - if the JSON file is effectively an array of test objects, I'm not sure how easy it is to parse that in all languages. Happy to investigate that though.

testing/storage/README.md
frankyn: @jadekler, thanks for putting this together.

IIRC, you mentioned that when Go tests read the JSON file the defined proto type can help make sure everyone is using the same type. 

Wdyt about the following:
1. Use the JSON file as the source of conformance tests and use the proto file as the expectation for types. Manual will be required.
1. Remove the Go generator of the JSON file. 
1. Add a test to verify JSON file type correctness when parsed into Go.

This could be added a presubmit test in PR changes to verify everyone is using the correct type when modifying the JSON file.

testing/storage/testdata/customer-supplied-encryption-key.json
jskeet: > The one tricky part thinking off the top of my head is working out how to handle the "repeated" aspect - if the JSON file is effectively an array of test objects, I'm not sure how easy it is to parse that in all languages. Happy to investigate that though.

Thinking about this, the simplest approach would be to have a single top level type which all JSON messages conform to, with repeated elements, e.g.

```proto
message TestFile {
  repeated SigningV4Test signing_v4_tests = 1;
}

message SigningV4Test {
   string description = 1;
   ...
}
```

That would lead to one "extra" level of indentation in the test files, but would avoid having to have the manifest, simplify the validation code, and remove the repeated part. I'm happy to create a proof of concept of this if it would be useful (although it would either be in C# or Java...)

```json
{
  "signingV4Tests": [
    {
      "description": "First test"
    },

    {
      "description": "First test"
    }
  ]
}
```

jskeet:
I have a sample PR in google-cloud-dotnet (where I know changing the Travis file won't cause any issues) showing a proposed plan - but it's not clear yet which repo this should end up. (Obviously not google-cloud-dotnet - but possibly a new repo just for conformance tests.)
See https://github.com/googleapis/google-cloud-dotnet/pull/2954 - comments welcome.

testing/storage/main.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+
+		{
+			Description: "POST for resumable uploads",
+			Bucket: "test-bucket",
+			Object: "test-object",
+			Method: "POST",
+			Expiration: 10,
+			Headers: map[string]*pb.ValueList{
+				"M-goog-resumable": {Values: []string{"start"}},
```
quartzmo: I think this should be `x-goog-resumable`.

frankyn: @quartzmo please ignore these conformance tests for now and instead use:
https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

quartzmo: @frankyn 👍  That's what I'm doing, but I did take a look and notice this error.

jadekler:
I'm closing this PR since discussion has concluded and we have a path forward. Thanks all for sharing opinions!

PR #283, opened 2019-05-21
Refactor conformance tests for Firestore to use json instead of textproto and remove binproto

BenWhitehead:
@frankyn @jadekler As discussed, I have updated the firestore conformance tests to be json rather than textproto.

Given the number of files poked during generation, it's probably easiest to read the commit message from each commit and go from there rather than trying to read all the file changes.


testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -2362,33 +2362,13 @@ func outputTestText(filename, comment string, t *tpb.Test) {
...
-		err2 := f.Close()
-		if err == nil {
-			err = err2
-		}
-	}()
-
-	fmt.Fprintln(f, "# DO NOT MODIFY. This file was generated by")
-	fmt.Fprintln(f, "# github.com/GoogleCloudPlatform/google-cloud-common/testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go.")
-	fmt.Fprintln(f)
-	doc.ToText(f, comment, "# ", "#    ", 80)
```
jadekler: Losing the comment is a bummer. Is there a way to preserve that?

BenWhitehead: The json file won't allow us to write a comment. We could add a property to the proto such that the comment is written into it along with the description.

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -17,16 +17,16 @@ package main
...
 	"os"
 	"path/filepath"
 	"strings"
 
-	tpb "github.com/googleapis/google-cloud-common/testing/firestore/genproto"
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	tspb "github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/golang/protobuf/ptypes/wrappers"
+	tpb "github.com/googleapis/google-cloud-common/testing/firestore/genproto"
```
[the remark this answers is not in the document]
BenWhitehead: Done in follow up commit

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -58,11 +58,11 @@ var (
 // others allow both.
 type writeTest struct {
 	// Textproto filename suffix.
-	suffix           string
+	suffix string
```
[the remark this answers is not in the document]
BenWhitehead: Done in follow up commit

BenWhitehead:
My plan was to squash once the review was complete as to not lose any context from the commits for the review.

Regarding the move to the conformance test repo, do we want to move everything over or just `testdata/test-suite.json` along with `proto/test.proto`. My understanding was that after we had generated the new json files, `generate-firestore-tests.go` would go away then in conformance-tests a new `validate` program would be added to ensure the json file(s) conform to the schema defined in `protos/test.proto`.

jadekler:
@BenWhitehead 

> My plan was to squash once the review was complete as to not lose any context from the commits for the review.

SG!

> Regarding the move to the conformance test repo, do we want to move everything over or just testdata/test-suite.json along with proto/test.proto. My understanding was that after we had generated the new json files, generate-firestore-tests.go would go away then in conformance-tests a new validate program would be added to ensure the json file(s) conform to the schema defined in protos/test.proto.

I loosely expect one PR here to delete `testing` and one PR in conformance-tests to add `firestore/v1` (incl whatever files you deem necessary). Does that seem right to you?

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
tseaver: +1 to adding a property for the comment.

BenWhitehead:
I've just pushed another commit that includes comments in the proto definition.

> I loosely expect one PR here to delete testing and one PR in conformance-tests to add firestore/v1 (incl whatever files you deem necessary). Does that seem right to you?

My plan was to get this PR merged for the format change, from conformance-tests point the the SHA where I'm going to copy the test-suite.json from, then have another PR here to delete this stuff and point to the new conformance-tests PR. This is so that anyone that is curious on the history or migration of these tests has some breadcrumbs they can follow.

jadekler:
> I've just pushed another commit that includes comments in the proto definition.
> 
> > I loosely expect one PR here to delete testing and one PR in conformance-tests to add firestore/v1 (incl whatever files you deem necessary). Does that seem right to you?
> 
> My plan was to get this PR merged for the format change, from conformance-tests point to the SHA where I'm going to copy the test-suite.json from, then have another PR here to delete this stuff and point to the new conformance-tests PR. This is so that anyone that is curious on the history or migration of these tests has some breadcrumbs they can follow.

SG!

BenWhitehead:
@jadekler I've squashed all the commits this is ready to merge when you get a chance.

jadekler:
Started playing with this a bit. Question: what's the purpose of `test_suite.json`, and how does it relate to all the other JSON files?

BenWhitehead:
`test-suite.json` is all of the individual tests in a single file. It matches up with the `TestSuite` defined in `tests.proto`. This was previously represented in the `test-suite.binproto` file.

jadekler:
> `test-suite.json` is all of the individual tests in a single file. It matches up with the `TestSuite` defined in `tests.proto`. This was previously represented in the `test-suite.binproto` file.
> 
> The existing java conformance tests for firestore use `test-suite.binproto`, I'll update to use `test-suite.json` once we move it over to conformance-tests.

Gotcha. I don't think we need that. @tseaver, do you think dynamic languages would need that ?

BenWhitehead:
The json file for storage is a single json file [storage/v1/v4_signatures.json](https://github.com/googleapis/conformance-tests/blob/master/storage/v1/v4_signatures.json). Which way do we want to use moving forward, I think we should be consistent regardless of what we pick.

jadekler:
> The json file for storage is a single json file [storage/v1/v4_signatures.json](https://github.com/googleapis/conformance-tests/blob/master/storage/v1/v4_signatures.json). Which way do we want to use moving forward, I think we should be consistent regardless of what we pick.

One of those is ~50 lines, the other is ~4000 lines. :) I am two thumbs down on trying to maintain a multi-thousand line JSON file.

Another note is that storage tests are nascent whereas firestore are mature: that is, storage only has one set of tests at the moment, whereas firestore has maaaany. The next set of storage tests are not guaranteed to live in `v4_signatures.json`. (the clue is in the name - that file only pertains to v4 sig feature)

----------------

Also: I'm working on converting our Go runner to take these .json tests just to verify we aren't goofing anything in the process. Might be good to do the same in Java (or language of your choice) to double check! :)

BenWhitehead:
> One of those is ~50 lines, the other is ~4000 lines. :) I am two thumbs down on trying to maintain a multi-thousand line JSON file.
Sounds fine to me. I do however feel strongly that we should only pick a single way to represent these. That way we set the precedent for any new projects looking at the existing definitions as examples.

> Also: I'm working on converting our Go runner to take these .json tests just to verify we aren't goofing anything in the process. Might be good to do the same in Java (or language of your choice) to double check! :)

I have a dev local branch in google-cloud-java that is updated to use the `test-suite.json` that is working. I'll have to update it use the individual files.

jadekler:
> > One of those is ~50 lines, the other is ~4000 lines. :) I am two thumbs down on trying to maintain a multi-thousand line JSON file.
> 
> Sounds fine to me. I do however feel strongly that we should only pick a single way to represent these. That way we set the precedent for any new projects looking at the existing definitions as examples.

Why do you feel strongly that way? It seems fairly easy to grok one test per file vs multiple tests per file, and there's not much difference in the runner implementation.

> > Also: I'm working on converting our Go runner to take these .json tests just to verify we aren't goofing anything in the process. Might be good to do the same in Java (or language of your choice) to double check! :)
> 
> I have a dev local branch in google-cloud-java that is updated to use the `test-suite.json` that is working. I'll have to update it use the individual files.

Sounds good!

jadekler:
Ignoring the one-test-per-file vs multiple-tests-per-file bikeshedding, this PR seems ready given that you've removed test-suite.json (did that happen? I assume so based on your comments but didn't check heh) and that you've tested against Java (thanks again for doing so).

blowmage:
Why is the `test-suite.binproto` file removed? From the PR title I expected maybe the `.textproto` files to be removed, but not `.binproto`.

BenWhitehead:
@jadekler I have not yet removed generation of `test-suite.json` I can go ahead and do that and update the PR. Do you want me to squash that update and push, or push a second commit and you squash from the github UI?

tseaver:
@jadekler 

>  do you think dynamic languages would need that ?

Python uses the individual `.textproto` files now, but Ruby doesn't have `.textproto` support, and so uses the single `.binproto` file.

BenWhitehead:
I've started looking into using the file-per-test in java and it's going to and up being a lot of work to write code to scan the classpath and find all the files. If we're okay keeping the big `test-suite.json` as a concatenation of all the small files I can do the work over in conformance-tests to generate it from each of the smaller files.

BenWhitehead:
@jadekler per our offline discussion I've removed `test-suite.json` this is ready to merge.

testing/firestore/cmd/generate-firestore-tests/generate-firestore-tests.go
```diff
@@ -607,22 +607,20 @@ func main() {
 	genDelete(suite)
 	genQuery(suite)
 	genListen(suite)
-	if err := writeProtoToFile(filepath.Join(*outputDir, "test-suite.binproto"), suite); err != nil {
```
blowmage: I would prefer to keep the .binproto file.

jadekler:
> @jadekler
> 
> > do you think dynamic languages would need that ?
> 
> Python uses the individual `.textproto` files now, but Ruby doesn't have `.textproto` support, and so uses the single `.binproto` file.

@blowmage @tseaver Sorry, we're still figuring out the kinks of the new https://github.com/googleapis/conformance-tests repo, including specification. Consequently we've not done a good job communicating upcoming changes. Let me try to fix that now:

- We plan on switching from .binproto and .textproto to just simple .json files. We expect all conformance tests to adopt this, and expect the firestore runners to be updated to use these json files.
- Have y'all adopted the [storage conformance tests](https://github.com/googleapis/conformance-tests/tree/master/storage/v1)? If so, I imagine that the firestore runners' file-walk logic to look very similar.
- Do either of y'all, or anyone else, foresee problems with this approach? Are there things in your language we aren't seeing?

jadekler:
Friendly ping @blowmage @tseaver. If you're thumbs up on this, we'll move forward with merge and continue migrating to https://github.com/googleapis/conformance-tests.

quartzmo:
> Have y'all adopted the storage conformance tests? 

@jadekler 

In google-cloud-storage (Ruby), we are using json conformance tests that I copied in before the files that you linked were available. The files we currently use are topically different but appear equivalent.

What is the mechanism for importing files from googleapis/conformance-tests? Do we simply manually copy them into our repo, or is there something more automated that we can do?

quartzmo:
@blowmage just told me that the import mechanism should be git submodule, is that correct?

quartzmo:
I see it now in the googleapis/conformance-tests README: 

> Generally, conformance tests are accessed as git submodules in the repositories.

quartzmo:
googleapis/google-cloud-ruby#3442

jadekler:
> What is the mechanism for importing files from googleapis/conformance-tests? Do we simply manually copy them into our repo, or is there something more automated that we can do?

Currently the easiest thing is to just copy paste. The recommended is git submodules, if it's feasible. @BenWhitehead is thinking about (and possibly working on?) other mechanisms.

BenWhitehead:
> > What is the mechanism for importing files from googleapis/conformance-tests? Do we simply manually copy them into our repo, or is there something more automated that we can do?
> 
> Currently the easiest thing is to just copy paste. The recommended is git submodules, if it's feasible. @BenWhitehead is thinking about (and possibly working on?) other mechanisms.

From my poking around, it looks like most of the languages are using copy-paste right now, the only language using git submodules I saw was dotnet. I've created https://github.com/googleapis/conformance-tests/issues/7 which we can use as a location to dicuss consumption of conformance test artifacts. Over on that issues a few of the recommendations I've heard during discussions are listed.

quartzmo:
This change will be moved/copied to the conformance test repo, correct?

If so, I'll try to import it into google-cloud-firestore (Ruby) from the conformance test repo using git submodules once it has been merged.

jskeet:
Likewise I'll try to do it for the dotnet repo. I'm also looking at moving the Bigtable conformance tests in there. (I'm assuming everyone else has the same tests as .NET...)

BenWhitehead:
@quartzmo Correct, I'm working on getting the tests moved over to googleapis/conformance-tests. I'll ping folks when I have a PR up.

PR #292, opened 2019-12-02
docs: add commits convention description in contributing/readme.md

paul1319:
Towards https://github.com/googleapis/google-cloud-python/issues/9897

### DIFF
--- a/contributing/readme.md
+++ b/contributing/readme.md
@@ -73,7 +73,7 @@ The following is a high-level overview of how to contribute to a Google Cloud cl
 
 1. Run the tests again.
 
-1. Commit your code. Take a moment to write a [good commit message](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). If you end up with several commits for one logical change, [squash](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) these commits for clarity.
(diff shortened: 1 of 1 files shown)


paul1319:
@IlyaFaer

tswast:
Let's link to https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type also, since it has the full set of commit types we support.

tswast:
Hmm... Except I don't see `chore` in there. @bcoe Do you have a better list of conventional commit types we use?

crwilcox:
I don't know if it was added, but I see it 

```
types other than fix: and feat: are allowed, for example @commitlint/config-conventional (based on the the Angular convention) recommends build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, and others.

```

bcoe:
@tswast @crwilcox I'm open to edits on conventionalcommits.org, which I help edit.

The motivation of the specification was to be slightly less prescriptive about the types that folks use, as the angular conventions were slightly more tailored to the angular project itself.

When I explain the specification to folks, I point out that the most important types are:

* `fix`: when you're fixing something.
* `feat`: when you're adding a feature.
* and an `!` point when you're breaking something (this was not part of the original angular specification, but is an important improvement to the specification in my opinion; folks kept forgetting to add `Breaking Change`, or typoing it, which would break our automation.

